PR #2, opened 2019-05-15
Optimized transpilation. Completed more of Send Event action type.

Aaronius:
## Description
I've improved the transpilation of the library modules to reduce the amount of emitted overhead code. I've completed the Send Event action type except for two things:

* It doesn't have UI tests. I want to see what it would take to set up TestCafe using the extension sandbox in order to test the extension's UI. I'll work on that next.
* The user will eventually be able to configure multiple Alloy instances from the extension configuration view and then select which instance should be used on the Send Event view. This experience is still being designed by UX, so right now the Send Event action type just assumes it should use the `alloy` instance.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-29012

## Motivation and Context
As an example, without these optimizations we get overhead like this:

```
var _createSendEvent = _interopRequireDefault(require("./createSendEvent"));
var _getInstance = _interopRequireDefault(require("../getInstance"));
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { "default": obj }; }
module.exports = (0, _createSendEvent["default"])(_getInstance["default"]);
```

With the optimization, it's reduced to this:

```
var _createSendEvent = require("./createSendEvent");
var _getInstance = require("../getInstance");
module.exports = (0, _createSendEvent.default)(_getInstance.default);
```

## How Has This Been Tested?
Automated and manual tests.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/src/lib/actions/sendEvent.js
+++ b/src/lib/actions/sendEvent.js
@@ -10,14 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-var getInstance = require("../getInstance.js");
+import createSendEvent from "./createSendEvent";
+import getInstance from "../getInstance";
 
-module.exports = function() {
-  var instance = getInstance("alloy");
-  instance("event", {
-    data: {
-      products: ["shirt", "shoes"],
-      total: [19.99]
-    }
-  });
-};
+module.exports = createSendEvent(getInstance);
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,11 @@ module.exports = {
     "no-param-reassign": ["error", { props: false }],
     "prettier/prettier": "error",
     "react/require-default-props": "off",
+    "jsx-a11y/label-has-associated-control": [ 2, {
+      "controlComponents": ["WrappedField"]
+    }],
+    // Has been deprecated in favor of label-has-associated-control
+    "jsx-a11y/label-has-for": "off",
     // Turning this off allows us to import devDependencies in our build tools.
     // We enable the rule in src/.eslintrc.js since that's the only place we
     // want to disallow importing extraneous dependencies.
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 dist
 .cache
 coverage
+.sandbox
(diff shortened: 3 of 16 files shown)


src/lib/actions/sendEvent.js
```diff
@@ -10,14 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-module.exports = function() {
-  var instance = getInstance("alloy");
-  instance("event", {
-    data: {
-      products: ["shirt", "shoes"],
-      total: [19.99]
-    }
-  });
-};
+module.exports = createSendEvent(getInstance);
```
Aaronius: Notice that while we can use `export default` elsewhere, we must use `module.exports` in the module that is exposed to Launch because Launch is expecting a CommonJS module.

Aaronius:
I'm closing this because it causes the following error while running unit tests:

```
Error: Rollup requires that your Babel configuration keeps ES6 module syntax intact. Unfortunately it looks like your configuration specifies a module transformer to replace ES6 modules with another module format. To continue you have to disable it.

Most commonly it's a CommonJS transform added by @babel/preset-env - in such case you should disable it by adding `modules: false` option to that preset (described in more detail here - https://github.com/rollup/rollup-plugin-babel#modules ).
    at error (/Users/aahardy/dev/rea/node_modules/rollup/dist/rollup.js:3866:30)
    at Object.error (/Users/aahardy/dev/rea/node_modules/rollup/dist/rollup.js:17317:17)
    at Object.error (/Users/aahardy/dev/rea/node_modules/rollup/dist/rollup.js:17750:31)
    at /Users/aahardy/dev/rea/node_modules/rollup-plugin-babel/dist/rollup-plugin-babel.cjs.js:73:9
    at /Users/aahardy/dev/rea/node_modules/rollup-plugin-babel/dist/rollup-plugin-babel.cjs.js:202:20
```

I think the best way to handle this is to not use ES6 modules in our library code.

PR #10, opened 2019-08-21
CORE-33751 Made name the primary identifier for each instance.

Aaronius:
## Description
I talked to Justin Grover about using name as the primary identifier for the instance rather than propertyId. In other words, if a user were to go to the Send Event action view, they would select an instance name from a dropdown to select the instance, whereas right now they're selecting a property ID. Selecting by name seems to make more sense to me. He agreed.

Also, I set up the passing of org ID into the `configure` command. I was doing some locally hardcoded stuff to make it work previously. I probably should have done this in a different PR, but alas, I got carried away.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33751

## Motivation and Context
The name, which is used for the global method on window, seems much more like a primary identifier of an instance that the property ID we pass through configuration.

## How Has This Been Tested?
Automated and manual

## Types of changes

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -74,7 +74,7 @@
         "$schema": "http://json-schema.org/draft-04/schema#",
         "type": "object",
         "properties": {
-          "propertyId": {
+          "name": {
             "type": "string",
             "minLength": 1
           },
@@ -88,7 +88,7 @@
           }
         },
         "required": [
-          "propertyId",
+          "name",
           "data"
         ],
         "additionalProperties": false
--- a/extension.json
+++ b/extension.json
@@ -54,7 +54,7 @@
             },
             "required": [
               "propertyId",
-              "instanceName"
+              "name"
             ],
             "additionalProperties": false
           }
@@ -74,7 +74,7 @@
         "$schema": "http://json-schema.org/draft-04/schema#",
         "type": "object",
         "properties": {
-          "name": {
+          "instanceName": {
             "type": "string",
             "minLength": 1
           },
@@ -88,7 +88,7 @@
           }
         },
         "required": [
-          "name",
+          "instanceName",
           "data"
         ],
         "additionalProperties": false
--- a/src/.eslintrc.js
+++ b/src/.eslintrc.js
@@ -3,5 +3,8 @@ module.exports = {
     // The eslintrc in the base directory turns this rule off.
     // We want to turn it back on for the src directory.
     "import/no-extraneous-dependencies": "error"
+  },
+  globals: {
+    _satellite: "readonly"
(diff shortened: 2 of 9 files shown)


Aaronius:
@jonsnyder Yeah, that's an interesting idea. I wouldn't want to maintain a list of window properties to validate against, but if we just checked their input against the properties on the extension view's `window` object, it would probably be a pretty accurate representation.

Aaronius:
I'll log a separate issue for that.

Aaronius:
One other way to solve this that I should mention is to add an `id` property inside the settings block for each instance and generate a unique value for it when the extension configuration is saved. Then the `id` property would become the primary identifier. When the user chooses an instance in the Send Event view, they would select the instance by name, but we would store `instanceId` in the action settings object rather than `instanceName`. This would would allow the user to be able to change the `name` of the instance at any time without breaking their rules and data elements, except for when they have custom code either inside or outside of Launch that references the global method we create on `window` (which I have no doubt will happen).

There's a downside to this. API users of Launch would need to generate this unique ID themselves, which is a little weird. I would suspect most of the time they wouldn't generate some UUID for `id`, but instead use whatever value they're already going to use for their `name`, which might reduce the value of `id` a bit?

Anyway, I'm not opposed to this approach, but I don't know if it's justified by the cost/benefit. I'm curious to hear what others think.

Aaronius:
Actually, the `id` property idea kind of gets more complicated. Let's say a user created an instance with an auto-generated `id of `oijwf823fl3` and with a `name` of `myalloy`. Then the user goes to create a Send Event action and selects `myalloy` for the instance they want to use and then saves the action. The user goes back to the extension configuration and deletes the `myalloy` instance. The user goes back to the action view. Because we would had only saved `instanceId` to the Send Event action settings and not `instanceName`, we no longer have the name value `myalloy` stored anywhere to show in the instance dropdown inside the Send Event action view (remember, the UX was to show the deleted instance in the dropdown). So we would either need to just show something like "deleted instance" in the instance dropdown, or we would need to save both `instanceId` AND `instanceName` into the Send Event action settings so it's always available regardless of whether the instance gets deleted from the extension configuration view. Storing `instanceName` into the action settings causes its own synchronization problems.

Anyway, I think it's reasonable to use `name` as the primary identifier.

PR #19, opened 2019-09-03
CORE-32727 Added support for event merge ID.

Aaronius:
## Description
Added an Event Merge ID data element type. I also added an Event Merge ID textfield to the Send Event view.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32727

## Motivation and Context
Users want to merge events.

## Types of changes

## Checklist:


extension.json
```diff
@@ -125,6 +129,26 @@
...
+            "minLength": 1
+          }
+        },
+        "required": [
+          "instanceName"
+        ],
+        "additionalProperties": false
+      },
+      "libPath": "dist/lib/dataElements/eventMergeId/index.js",
+      "viewPath": "dataElements/eventMergeId.html"
```
Aaronius: I want to call attention to something here. The ECID and Event Merge ID views are identical. They both just ask for the instance to use. We could have a single HTML file called something like `instanceNameOnly.html` and then set the `viewPath` for both data elements types to `dataElements/instanceNameOnly.html`. This is totally a legitimate option.

For some reason, I felt that it would be preferable to give each data element type its own HTML file. So I created `ecid.html` and `eventMergeId.html`. I think I maybe felt that way because having `dataElements/instanceNameOnly.html` might be slightly misleading since "instance name only" isn't a data element type....the name just represents that it's a view that's only asking for an instance name. Perhaps we could name it something like `ecidAndEventMergeId.html` (which isn't really scalable if we have more data element types that use the same name).

Having both `ecid.html` and `eventMergeId.html` led to some other things. First, they couldn't both point to the same JavaScript file because of https://github.com/adobe/reactor-extension-alloy/pull/19/files#diff-2d877a9354b41b53de8b107b77570012R16 (see https://github.com/adobe/reactor-extension-alloy/pull/19/files#diff-2d877a9354b41b53de8b107b77570012R13-R17 for the workaround). Second, I felt obligated to run functional tests on both `ecid.html` and `eventMergeId.html`, even though they both behave exactly the same. Third, because of the Parcel bug I just mentioned and its related workaround, the JavaScript bundle used for `ecid.html` is different that the one used for `eventMergeId.html`, even though the code is exactly the same.

Anyway, now I'm questioning whether I actually made the right choice and considering going back to the `instanceNameOnly.html` idea. Thoughts?

Aaronius:
Don't review this PR in its entirety yet. I'm still debating what I want to do about https://github.com/adobe/reactor-extension-alloy/pull/19#discussion_r320470287

Aaronius:
Yes, probably, but I'm not very creative. 😢  Here were some things I thought about:

* `instanceNameOnly.html` I actually chose this one first and then talked myself out of it. I think the reason I decided not to go with this is because the data element isn't intended to return an "instance name only"; it's just that the view asks for an instance name only, and I thought that might be confusing. In retrospect, I don't know that it's any more confusing than `minimal.html`.
* `ecidAndEventMergeId.html` This is very explicit, but not very scalable if/when we add more data element types that use the same view, since the name will just grow longer and longer.

Do you prefer one of those or do you have other ideas? I could probably get behind `instanceNameOnly.html`.

Aaronius:
@jfkhoury @owenleej Any feedback on https://github.com/adobe/reactor-extension-alloy/pull/19#issuecomment-527956115?

jfkhoury:
> Yes, probably, but I'm not very creative. 😢 Here were some things I thought about:
> 
> * `instanceNameOnly.html` I actually chose this one first and then talked myself out of it. I think the reason I decided not to go with this is because the data element isn't intended to return an "instance name only"; it's just that the view asks for an instance name only, and I thought that might be confusing. In retrospect, I don't know that it's any more confusing than `minimal.html`.
> * `ecidAndEventMergeId.html` This is very explicit, but not very scalable if/when we add more data element types that use the same view, since the name will just grow longer and longer.
> 
> Do you prefer one of those or do you have other ideas? I could probably get behind `instanceNameOnly.html`.

I can get behind `instanceNameOnly.html` as well :)

Aaronius:
Done. Thanks!

PR #20, opened 2019-09-05
CORE-33720 Added Set Opt-In Preferences action type.

Aaronius:
## Description

Adds Set Opt-In Preferences action type. I also made a change across several of the views where, if a user is creating an action or data element (instead of editing an existing one), we now auto-select the first configured instance on their behalf. This makes for fewer user clicks and also ensures that an instance will always be selected, which simplifies validation.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33720

## Motivation and Context
Customers need to convey user opt-in preferences to Alloy.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/actions/setOptInPreferences.jsx
+++ b/src/view/actions/setOptInPreferences.jsx
@@ -1,0 +1,109 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import "regenerator-runtime"; // needed for some of react-spectrum
+import React from "react";
+import Alert from "@react/react-spectrum/Alert";
+import Select from "@react/react-spectrum/Select";
+import RadioGroup from "@react/react-spectrum/RadioGroup";
+import Radio from "@react/react-spectrum/Radio";
+import "@react/react-spectrum/Form"; // needed for spectrum form styles
+import render from "../render";
+import WrappedField from "../components/wrappedField";
+import ExtensionView from "../components/extensionView";
+import getInstanceOptions from "../utils/getInstanceOptions";
+import "./setOptInPreferences.styl";
+
+const purposesEnum = {
+  ALL: "all",
+  NONE: "none"
+};
+
+const getInitialValues = initInfo => {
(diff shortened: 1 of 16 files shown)


src/view/actions/setOptInPreferences.jsx
```diff
@@ -0,0 +1,109 @@
...
+import RadioGroup from "@react/react-spectrum/RadioGroup";
+import Radio from "@react/react-spectrum/Radio";
+import "@react/react-spectrum/Form"; // needed for spectrum form styles
+import render from "../render";
+import WrappedField from "../components/wrappedField";
+import ExtensionView from "../components/extensionView";
+import getInstanceOptions from "../utils/getInstanceOptions";
+import "./setOptInPreferences.styl";
+
+const purposesEnum = {
```
Aaronius: I called this `purposesEnum` instead of `purposes` because `purposes` would conflict with other `purposes` variables and it made sense to make this one more explicit. To be consistent, I also renamed `contextGranularity` to `contextGranularityEnum` inside the extension configuration view.

src/view/actions/setOptInPreferences.jsx
```diff
@@ -0,0 +1,109 @@
...
+  ALL: "all",
+  NONE: "none"
+};
+
+const getInitialValues = initInfo => {
+  const {
+    instanceName = initInfo.extensionSettings.instances[0].name,
```
Aaronius: This is where the magic happens of auto-selecting the first instance. I've made the same kind of change to the other views. Note that we can make the assumption that there will always be at least one instance in the `instances` array because the extension configuration schema enforces at least one instance to always be configured.

src/view/dataElements/instanceNameOnly.jsx
```diff
@@ -40,16 +35,11 @@ const getSettings = values => {
   };
 };
 
-const validationSchema = object().shape({
-  instanceName: string().required("Please specify an instance")
-});
-
```
Aaronius: Notice that since we're auto-selecting an instance name now, the instance name field will always have an instance selected, which negates the need for any validation on the field.

src/view/actions/setOptInPreferences.jsx
jfkhoury: I love this change!!!

owenleej:
@Aaronius @jfkhoury When we added opt-in to the ECID launch extension it initially had actions for doing opt-in stuff and we removed them for some reason. Do you remember what the reason was and if it applies in any way to this extension?

Aaronius:
@owenleej Thanks for bringing that up. If I remember correctly, they were removed from the ECID extension because we assumed customers would always be setting opt-in preferences from on-page code. @jfkhoury, do you know if this is still the case? If it is, maybe this action is unnecessary?

Aaronius:
I talked this over with Justin and we agree that it shouldn't be an action in the Launch extension. I'm going to make a separate PR that only contains the stuff from this PR that's still pertinent (e.g., auto-selection of the first instance name)

PR #24, opened 2019-09-12
CORE-33593 Updated for the xdm/data options refactor.

Aaronius:
## Description
Updated for the xdm/data options refactor. `data` changed to `xdm`. `xdm` is now optional (`data` used to be required). The event merge ID field is removed from the Send Event view, since the event merge ID will be a part of the XDM object the customer builds.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33593

## Motivation and Context
Needs to work with the latest Alloy.

## Types of changes

## Checklist:

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -90,18 +90,13 @@
             "type": "boolean",
             "minLength": 1
           },
-          "data": {
+          "xdm": {
             "type": "string",
             "pattern": "^%([^%]+)%$"
-          },
-          "eventMergeId": {
-            "type": "string",
-            "minLength": 1
           }
         },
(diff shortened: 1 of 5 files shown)


Aaronius:
@jonsnyder The `eventMergeId` used to be a top-level property in the event object when sent to Konductor, similar to `web`, `device`, etc., but now it's not. Since `eventMergeId` will be a part of the customer's XDM schema, they'll just pass it in their `xdm` block (or if they pass it somewhere in `data`, then they'll map whatever key they use to `eventMergeId` within the mapping service).

PR #25, opened 2019-09-16
CORE-33720 Added Set Opt-In Preferences action type.

Aaronius:
## Description
After talking with Alpha customers, it does sounds like a Set Opt-In Preferences action type is useful, but they would like to be able to set opt-in preferences both statically (e.g., "all" and "none" checkboxes) as well as via a data element. This PR provides for both.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33720

image

## Types of changes

## Checklist:

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -102,6 +102,42 @@
       },
       "libPath": "dist/lib/actions/sendEvent/index.js",
       "viewPath": "actions/sendEvent.html"
+    },
+    {
+      "displayName": "Set Opt-In Preferences",
+      "name": "set-opt-in-preferences",
+      "schema": {
+        "$schema": "http://json-schema.org/draft-04/schema#",
+        "type": "object",
+        "properties": {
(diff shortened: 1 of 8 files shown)


Aaronius:
@jonsnyder I totally agree. I was thinking of waiting until later to figure out what goes in tooltips vs documentation, but maybe I'll take a stab at it soon.

@jfkhoury From the feedback, here's why I think the checkboxes still might make sense (I think @jonsnyder and I had a conversation about this in the hall too). If my website has a dialog that allows the user you opt into everything or opt into nothing, then as a Launch user I could set up a rule with an event that detects when the user clicks the "opt into everything button" and an action that sets the opt-in preferences to "all" (as a Launch user I would check the checkbox). I could then make a separate rule for when the user clicks the "opt into nothing button" and an action that sets the opt-in preferences to "none". Does that make sense?

PR #27, opened 2019-10-02
CORE-34539 Fixed validation for ID sync container ID.

Aaronius:
## Description
Fixed validation for ID sync container ID. Now the ID sync container ID field will only show up when the ID Synchronization Enabled checkbox is checked, since that really only when it matters. Acceptable values are an integer over 0, a single data element token, or an empty string. If the value is an empty string or the ID Synchronization Enabled checkbox is not checked, the ID sync container ID will not be added to the settings object.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-34539

## Motivation and Context
Proper validation

## Types of changes

## Checklist:

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -42,8 +42,16 @@
                 "type": "boolean"
               },
               "idSyncContainerId": {
-                "type": "number",
-                "minimum": 0
+                "oneOf": [
+                  {
+                    "type": "integer",
+                    "minimum": 0
+                  },
+                  {
+                    "type": "string",
+                    "pattern": "^%([^%]+)%$"
+                  }
+                ]
               },
               "destinationsEnabled": {
                 "type": "boolean"
--- a/extension.json
+++ b/extension.json
@@ -49,7 +49,7 @@
                   },
                   {
                     "type": "string",
-                    "pattern": "^%([^%]+)%$"
+                    "pattern": "^%[^%]+%$"
                   }
                 ]
               },
@@ -100,7 +100,7 @@
           },
           "xdm": {
             "type": "string",
-            "pattern": "^%([^%]+)%$"
+            "pattern": "^%[^%]+%$"
           }
         },
         "required": [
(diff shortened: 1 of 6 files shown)


extension.json
```diff
@@ -42,8 +42,16 @@
...
-                "type": "number",
-                "minimum": 0
+                "oneOf": [
+                  {
+                    "type": "integer",
+                    "minimum": 0
+                  },
+                  {
+                    "type": "string",
+                    "pattern": "^%([^%]+)%$"
```
jonsnyder: are the parentheses necessary here?

test/functional/configuration/configuration.spec.js
```diff
@@ -286,22 +287,54 @@ test("shows errors for duplicate names", async t => {
   await instances[1].nameField.expectError(t);
 });
 
-test("shows error for negative ID sync container ID", async t => {
+test("shows error for non-negative number for ID sync container ID", async t => {
```
jonsnyder: non-negative?

test/functional/configuration/configuration.spec.js
```diff
@@ -286,22 +287,54 @@ test("shows errors for duplicate names", async t => {
...
+  await extensionViewController.expectIsNotValid(t);
+  await instances[0].idSyncContainerIdField.expectError(t);
+});
+
+test("does not show error for a data element for ID sync container ID", async t => {
+  await extensionViewController.init(t, {});
+  await instances[0].propertyIdField.typeText(t, "PR123");
+  await instances[0].idSyncContainerIdField.typeText(t, "%123foo%");
+  await extensionViewController.expectIsValid(t);
+});
```
jonsnyder: Do you want to test a multi-element data element?

extension.json
Aaronius: No! Thanks for spotting that.

test/functional/configuration/configuration.spec.js
Aaronius: Yeah, that should be removed. Thanks.

test/functional/configuration/configuration.spec.js
Aaronius: It's a good idea. I'll do it.

jonsnyder:
👍

PR #28, opened 2019-10-04
CORE-35804 Added org ID to instance configuration

Aaronius:
## Description
Added org ID to instance configuration.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-35804

## Motivation and Context
Two instances on the same page need to have unique org IDs. Alloy validates that this is the case.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/components/extensionView.jsx
+++ b/src/view/components/extensionView.jsx
@@ -26,23 +26,37 @@ const ExtensionView = ({
   validationSchema,
   render
 }) => {
-  const [{ initialized, initialValues, initInfo }, setState] = useState({
+  const [
+    { initialized, initialValues, initInfo },
+    setInitializationValues
+  ] = useState({
     initialized: false,
     initialValues: {},
     initInfo: {}
   });
 
   useEffect(() => {
+    // This effect function only runs once, so while our registered init
+    // function updates the component's state using setInitializationValues,
+    // the getSettings closure wouldn't be able to access the updated component
+    // state value. For this reason, we keep this local effectInitInfo
+    // so that getSettings() can access the initInfo value after it comes
+    // through init().
+    let effectInitInfo;
     window.extensionBridge.register({
       init: _initInfo => {
-        setState({
+        effectInitInfo = _initInfo;
+        setInitializationValues({
           initialized: true,
-          initialValues: getInitialValues(_initInfo),
+          initialValues: getInitialValues({ initInfo: _initInfo }),
           initInfo: _initInfo
         });
       },
       getSettings: () => {
-        return getSettings(formikProps.values);
+        return getSettings({
+          values: formikProps.values,
+          initInfo: effectInitInfo
+        });
       },
       validate: () => {
         // The docs say that the promise submitForm returns
--- a/src/view/configuration/configuration.jsx
+++ b/src/view/configuration/configuration.jsx
@@ -39,10 +39,11 @@ const contextGranularityEnum = {
 };
 const contextOptions = ["web", "device", "environment", "placeContext"];
 
-const instanceDefaults = {
+const getInstanceDefaults = initInfo => ({
   name: "alloy",
   propertyId: "",
-  edgeDomain: "",
+  imsOrgId: initInfo.company.orgId,
+  edgeDomain: "alpha.konductor.adobedc.net",
   errorsEnabled: true,
   optInEnabled: false,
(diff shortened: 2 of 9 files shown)


Aaronius:
Both the IMS org ID and the Edge Domain fields have a default value (the org ID is whatever the user's active org ID is, the edge domain is what we have hard-coded inside Alloy). I discussed the how this should be handled as far as UX goes with Alan Wilson and he thought (and I agree) that we should populate both textfields with the default value, make the fields required, and provide a button that allows the user to restore the default value in the textfield. This is what I've done in this PR.

Upon save, if the values for those fields match the default values, I don't store them inside the settings object, because it's useless to do so, since the user's org can automatically be pulled from `_satellite.company.orgId` and the default `edgeDomain` is hard-coded inside Alloy.

src/view/configuration/configuration.jsx
```diff
@@ -39,10 +39,11 @@ const contextGranularityEnum = {
 };
 const contextOptions = ["web", "device", "environment", "placeContext"];
 
-const instanceDefaults = {
+const getInstanceDefaults = initInfo => ({
```
Aaronius: We now need `initInfo` when getting the default values for an instance because that's where the user's active org ID comes from.

Aaronius:
Don't review this quite yet. I think I'm going to make a tweak.

src/view/components/extensionView.jsx
```diff
@@ -26,23 +26,37 @@ const ExtensionView = ({
...
   });
 
   useEffect(() => {
+    // This effect function only runs once, so while our registered init
+    // function updates the component's state using setInitializationValues,
+    // the getSettings closure wouldn't be able to access the updated component
+    // state value. For this reason, we keep this local effectInitInfo
+    // so that getSettings() can access the initInfo value after it comes
+    // through init().
+    let effectInitInfo;
```
Aaronius: This is a weird conundrum. There might be a better way to structure this component/extensionbridge relationship, but I would need to give it some deep thought.

src/view/configuration/configuration.jsx
```diff
@@ -75,23 +77,25 @@ const getInitialValues = initInfo => {
...
 
-const getSettings = values => {
+const getSettings = ({ values, initInfo }) => {
+  const instanceDefaults = getInstanceDefaults(initInfo);
   return {
     instances: values.instances.map(instance => {
       const trimmedInstance = {
-        name: instance.name,
-        propertyId: instance.propertyId
+        name: instance.name
```
Aaronius: `name` is the only property that needs to be saved to the settings object when it matches the form default value (`alloy`). We may even want to remove the default form value of "alloy" and just have the user always provide a name...I'm not sure. Considering "alloy" is the default in the base code in our alloy documentation, I think it probably makes sense to keep it as a default here as well.

test/functional/configuration/configuration.spec.js
```diff
@@ -252,29 +271,47 @@ test("returns full valid settings", async t => {
...
+  await instances[0].propertyIdField.typeText(t, "PR123");
+  await addInstanceButton.click(t);
+  await instances[1].propertyIdField.typeText(t, "PR456");
+  // We'll expand the first instance before we validate to test that
+  // validation expands the invalid instance (in this case, the second one)
+  // Even though both accordion header labels are "alloy", this
+  // will select the first one.
+  await accordion.clickHeader(t, "ALLOY");
+  await extensionViewController.expectIsNotValid(t);
+  await instances[1].nameField.expectError(t);
```
Aaronius: I just moved this test. It used to be below "shows errors for duplicate property IDs".

src/view/configuration/configuration.jsx
wgamboa: Are the parens needed here? I may be wrong, but I think it would return {} normally.

Aaronius: Without the parenthesis, the browser sees the curly braces as part of a regular function (e.g., `(initInfo) => { ... }` and will throw an error when it parses `name: "alloy"`. See the Advanced Syntax section here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

wgamboa: Oh, thanks. I didn't notice that

PR #29, opened 2019-10-07
CORE-33760 CORE-34548 Improved name validation.

Aaronius:
## Description
Added validation that ensures that the name has at least one character other than a digit and tries to ensure that it does not match a property already on window. It checks the name against the extension view's window, which isn't necessarily an accurate representation of the native window properties the user will encounter on the customer's website, but it should be a pretty good litmus test.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-34548
https://jira.corp.adobe.com/browse/CORE-33760

## Motivation and Context
Prevent errors due to problematic names.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/configuration/configuration.jsx
+++ b/src/view/configuration/configuration.jsx
@@ -150,7 +150,20 @@ const validationSchema = object()
   .shape({
     instances: array().of(
       object().shape({
-        name: string().required("Please specify a name."),
+        name: string()
+          .required("Please specify a name.")
+          // Under strict mode, setting window["123"], where the key is all
+          // digits, throws a "Failed to set an indexed property on 'Window'" error.
+          // This regex ensure there's at least one non-digit.
+          .matches(/\D+/, "Please provide a non-numeric name.")
+          .test({
+            name: "notWindowPropertyName",
+            message:
+              "Please provide a name that does not conflict with a property already found on the window object.",
+            test(value) {
+              return !(value in window);
+            }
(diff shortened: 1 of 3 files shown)


src/view/configuration/configuration.jsx
```diff
@@ -150,7 +150,20 @@ const validationSchema = object()
   .shape({
     instances: array().of(
       object().shape({
-        name: string().required("Please specify a name."),
+        name: string()
+          .required("Please specify a name.")
+          // Under strict mode, setting window["123"], where the key is all
+          // digits, throws a "Failed to set an indexed property on 'Window'" error.
+          // This regex ensure there's at least one non-digit.
```
jfkhoury: Good catch! Should we add this to the docs?

Aaronius: Yes. Done! https://github.com/AdobeDocs/alloy-docs/pull/54

test/functional/configuration/configuration.spec.js
```diff
@@ -296,7 +296,23 @@ test("shows errors for duplicate names", async t => {
   await instances[1].nameField.expectError(t);
 });
 
-test("shows errors for duplicate property IDs", async t => {
+test("shows error for numeric name", async t => {
```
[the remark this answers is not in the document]
Aaronius: wtf. I'm dumb. Thanks.

PR #30, opened 2019-10-07
CORE-33723 Display a warning when name changes on existing configuration.

Aaronius:
## Description
Display a warning when name changes on existing configuration.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33723

## Motivation and Context
Resources can break when an instance name is changed.

## Types of changes

## Checklist:

### DIFF
--- a/test/functional/configuration/configuration.spec.js
+++ b/test/functional/configuration/configuration.spec.js
@@ -29,6 +29,7 @@ const instances = [];
 for (let i = 0; i < 2; i += 1) {
   instances.push({
     nameField: spectrum.textfield(Selector(`[name='instances.${i}.name']`)),
+    nameChangeAlert: spectrum.alert(Selector("#nameChangeAlert")),
     propertyIdField: spectrum.textfield(
       Selector(`[name='instances.${i}.propertyId']`)
     ),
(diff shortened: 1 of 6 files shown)


test/functional/configuration/configuration.spec.js
```diff
@@ -312,6 +313,30 @@ test("shows error for name that matches key on window", async t => {
...
+    })
+  );
+  await instances[0].nameField.typeText(t, "123");
+  await instances[0].nameChangeAlert.expectExists(t);
+});
+
+test("does not show a warning when name is changed on new configuration", async t => {
+  await extensionViewController.init(t, defaultInitInfo);
+  await instances[0].nameField.typeText(t, "123");
+  await instances[0].nameChangeAlert.expectNotExists(t);
```
wgamboa: How does it know how long it has to wait before deciding that it doesn't exist?

Aaronius: TestCafe has a built-in polling mechanism with a timeout: https://devexpress.github.io/testcafe/documentation/test-api/built-in-waiting-mechanisms.html

Aaronius: In other words, if the element exists, it keeps polling until the element either doesn't exist (therefore the assertion passes) or the timeout is reached (therefore the assertion fails).

wgamboa: So if the element doesn't exist from the get go, it automatically passes? I guess there's no way to know if it appears later on.

Aaronius: Yeah, that's right.

PR #31, opened 2019-10-16
CORE-33204 Added tooltips and aligned UI more closely with mocks.

Aaronius:
## Description
It was pretty closely aligned with the mocks. I just needed to change a couple things and add tooltips. I'll be reviewing everything with UX again to make sure we're still on the same page now that the extension is more complete.

Most of these changes were for tooltips. I'll point out the specific things that are unrelated to tooltips.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33204

## Motivation and Context
Fulfill UX designs. Make the extension easier to use.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/configuration/configuration.jsx
+++ b/src/view/configuration/configuration.jsx
@@ -22,6 +22,7 @@ import Button from "@react/react-spectrum/Button";
 import Alert from "@react/react-spectrum/Alert";
 import ModalTrigger from "@react/react-spectrum/ModalTrigger";
 import Dialog from "@react/react-spectrum/Dialog";
+import FieldLabel from "@react/react-spectrum/FieldLabel";
 import Delete from "@react/react-spectrum/Icon/Delete";
 import { Accordion, AccordionItem } from "@react/react-spectrum/Accordion";
 import CheckboxList from "../components/checkboxList";
@@ -30,6 +31,7 @@ import render from "../render";
 import WrappedField from "../components/wrappedField";
 import ExtensionView from "../components/extensionView";
 import EditorButton from "../components/editorButton";
+import InfoTipLayout from "../components/infoTipLayout";
 import copyPropertiesIfNotDefault from "./utils/copyPropertiesIfNotDefault";
 import singleDataElementRegex from "../constants/singleDataElementRegex";
 import "./configuration.styl";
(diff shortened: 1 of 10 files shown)


src/view/configuration/configuration.jsx
```diff
@@ -233,7 +235,10 @@ const validationSchema = object()
   });
 
 const Configuration = () => {
-  const [selectedAccordionIndex, setSelectedAccordionIndex] = useState(0);
+  const [selectedAccordionIndex, setSelectedAccordionIndex] = useState();
+  const [isFirstExtensionViewRender, setIsFirstExtensionViewRender] = useState(
+    true
+  );
```
Aaronius: This code is not related to tooltips. It helps make sure that the accordion is collapsed when a configuration is loaded that contains multiple instances (per UX design).

src/view/configuration/configuration.jsx
```diff
@@ -516,34 +539,39 @@ const Configuration = () => {
...
+                                icon={<Delete />}
+                                variant="action"
+                                disabled={values.instances.length === 1}
+                              />
+                              {values.instances.length === 1 ? (
+                                <span className="Note u-gapLeft">
+                                  You must have at least one instance to use
+                                  this extension.
+                                </span>
+                              ) : null}
```
Aaronius: This delete button and note is also a change unrelated to tooltips. Specifically, we now always show the delete button, but unless there are multiple instances, the button is disabled at the user sees the "You must have at least one instance to use this extension." note.

src/view/utils.styl
```diff
@@ -39,3 +39,11 @@ $gap = 8px;
 .u-fieldLong {
   width: 400px !important;
 }
+
+.u-flex {
```
artur-ciocanu: Any reason we have to use `!important` here and below?

Aaronius: Yes, these are "utility classes". When you use them, you want them to override any conflicting styles the element may have received. If you didn't want them to override conflicting styles, you'd just leave the utility class off the element.

Related reading:

https://davidtheclark.com/on-utility-classes/
https://css-tricks.com/when-using-important-is-the-right-choice/

Precedents in other common CSS libraries: 

https://github.com/suitcss/utils-align/blob/master/lib/align.css
https://github.com/zurb/foundation-sites/blob/develop/scss/prototype/_text-utilities.scss
https://github.com/uikit/uikit/blob/0c3d7058f0f59ca2bcfe300a47edb14d5e10229b/src/less/components/margin.less
https://github.com/twbs/bootstrap/blob/ffaad0a819c6cf1e21d9a9d313673f9d7a260584/scss/mixins/_float.scss

Aaronius: I added a small comment to the top of the file in case others have the same question.

PR #32, opened 2019-10-23
Rename propertyId to configId.

jfkhoury:
There is no concept of Property ID as of yet. It's really the config ID that will be provided when you configure your property in the Configuration Service UI.

https://jira.corp.adobe.com/browse/CORE-35789

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x ] I've updated the schema in extension.json or no changes are necessary.

### DIFF
--- a/src/lib/runAlloy.js
+++ b/src/lib/runAlloy.js
@@ -2242,9 +2242,9 @@ governing permissions and limitations under the License.
         return function (config) {
           var componentRegistry = createComponentRegistry();
           var imsOrgId = config.imsOrgId,
-            propertyId = config.propertyId,
+            configId = config.configId,
             cookieDomain = config.cookieDomain;
-          var cookieName = ALLOY_COOKIE_NAME + "_" + propertyId;
+          var cookieName = ALLOY_COOKIE_NAME + "_" + configId;
           var cookieProxy = createCookieProxy(cookieName, ALLOY_COOKIE_TTL_IN_DAYS, cookieDomain || memoizedGetTopLevelDomain(window, cookie)); // TODO: Should this storage be namespaced by property ID or org ID?
 
           var storage = createNamespacedStorage(imsOrgId);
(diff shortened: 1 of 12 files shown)


src/lib/runAlloy.js
```diff
@@ -2242,9 +2242,9 @@ governing permissions and limitations under the License.
         return function (config) {
           var componentRegistry = createComponentRegistry();
           var imsOrgId = config.imsOrgId,
-            propertyId = config.propertyId,
+            configId = config.configId,
             cookieDomain = config.cookieDomain;
-          var cookieName = ALLOY_COOKIE_NAME + "_" + propertyId;
+          var cookieName = ALLOY_COOKIE_NAME + "_" + configId;
```
jfkhoury: Hey @Aaronius, I think we need to update this to use the new Cookie naming structure?

src/view/configuration/configuration.jsx
```diff
@@ -44,7 +44,7 @@ const contextOptions = ["web", "device", "environment", "placeContext"];
 
 const getInstanceDefaults = initInfo => ({
   name: "alloy",
-  propertyId: "",
+  configId: "",
   imsOrgId: initInfo.company.orgId,
   edgeDomain: "alpha.konductor.adobedc.net",
```
jfkhoury: We are currently hitting `beta.adobedc.net`. Should we update it here now, or wait for the 3rd party domain changes that are coming?

src/lib/runAlloy.js
Aaronius: The contents of `runAlloy.js` is currently manually copied in from the alloy build. In other words, you should run `npm run build:prod` in Alloy, then copy the contents of `dist/reactor/alloy.js` from Alloy and paste it into this `src/lib/runAlloy.js` file (see the comments for where to paste it). Is this answering your question?

src/view/configuration/configuration.jsx
Aaronius: Changing it here would be good.

Aaronius:
👍

PR #36, opened 2019-11-13
Rename imsOrgId to orgId.

jfkhoury:
## Description

ims is an internal concept. Please don't merge until the change is made on Konductor.

## Types of changes

- [ x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ x ] I have updated the documentation accordingly.
- [ x ] All tests pass and I've made any necessary test changes.
- [ x ] I've updated the schema in extension.json or no changes are necessary.

### DIFF
--- a/src/view/configuration/configuration.jsx
+++ b/src/view/configuration/configuration.jsx
@@ -45,7 +45,7 @@ const contextOptions = ["web", "device", "environment", "placeContext"];
 const getInstanceDefaults = initInfo => ({
   name: "alloy",
   configId: "",
-  imsOrgId: initInfo.company.orgId,
+  orgId: initInfo.company.orgId,
   edgeDomain: "beta.adobedc.net",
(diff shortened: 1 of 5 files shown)


src/view/configuration/configuration.jsx
```diff
@@ -227,11 +227,11 @@ const validationSchema = object()
   // TestCafe doesn't allow this to be an arrow function because of
   // how it scopes "this".
   // eslint-disable-next-line func-names
-  .test("uniqueImsOrgId", function(settings) {
+  .test("uniqueorgId", function(settings) {
```
Aaronius: I don't remember how the name actually provides value here, but I think we should camelcase it (`uniqueOrgId`).

jfkhoury: Will do, and will update `runAlloy.js`. Thanks!

PR #43, opened 2019-11-22
Add predefined event types to sendEvent action

jonsnyder:
## Description

This adds the predefined xdm event types to the sendEvent action.  I'm using a modified comboBox that will work with react.  In standup we talked about creating radio buttons for "predefined" versus "custom" values for this field, so that we could support friendly names for the events.  The problem is that there aren't any standard friendly names defined in XDM.  I could just add spaces to the values, but I feel like that would be more confusing.

Additionally, I re-ordered the fields to be: type, xdm, mergeId, viewStart.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-37378

## Motivation and Context

https://github.com/adobe/xdm/blob/master/docs/reference/data/time-series.schema.md

## Types of changes

## Checklist:


src/view/components/controlledCombo.jsx
```diff
@@ -0,0 +1,36 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import React from "react";
+import ComboBox from "@react/react-spectrum/ComboBox";
+
+// Use this component instead of a regular ComboBox to avoid the bug where you have to click
+// twice to select an option
```
Aaronius: I think it would be good to add the issue number to the comment so that we can remove this module when it's fixed.

Aaronius: How do I reproduce the problem? I tested with the following code and didn't have to click twice to select an option:

```
export default class ControlledCombo extends React.Component {
  // state = {
  //   showMenu: false
  // };

  render() {
    const {
      props,
      // state: { showMenu }
    } = this;
    return (
      <ComboBox
        // showMenu={showMenu}
        // onMenuToggle={menuState => this.setState({ showMenu: menuState })}
        {...props}
      />
    );
  }
}
```

src/view/actions/sendEvent.jsx
```diff
@@ -92,44 +117,37 @@ const SendEvent = () => {
...
-                  label="Occurs at the start of a view"
-                />
-              </InfoTipLayout>
-            </div>
             <div className="u-gapTop">
               <InfoTipLayout
-                tip="Please specify a data element that will return a JavaScript
-                  object in XDM format. This object will be sent to the Adobe
-                  Experience Platform."
+                tip="The type of the experience event.  Choose a predefined type or create
```
Aaronius: Remove the extra space between sentences please.

src/view/actions/sendEvent.jsx
```diff
@@ -92,44 +117,37 @@ const SendEvent = () => {
...
+                  your own. This will be added to the XDM object as the field `eventType`."
```
Aaronius: I think users are probably not familiar with backticks. Maybe just `eventType` or `"eventType"`? I'm going to ask on Slack what others think, because this has come up before.

src/view/components/controlledCombo.jsx
jonsnyder: When I used the ComboBox directly, I could reproduce the double clicking bug.  When I looked at the react-spectrum storybook code, I noticed it created a ControlledCombo control, so I figured the showMenu state was necessary.

Aaronius: I just tried again using the ComboBox directly and I'm not seeing any issues. Am I missing something?

My guess with the react-spectrum storybook is that they want to demonstrate that it's possible to control the menu state, but it probably isn't necessary.

src/view/actions/sendEvent.jsx
Aaronius: An update: I asked in Slack and it was pretty inconclusive. It looks like we're using backticks in a few other places and were put there by more than one other developer, so I think we should go ahead with it unless you feel otherwise.

src/view/components/controlledCombo.jsx
jonsnyder: I tried it again and it worked without the controlledCombo.  Thanks for looking into this.  I thought it was odd to have to do that.

PR #45, opened 2019-11-25
Adding onBeforeEventSend

jileeshs:
onBeforeEventSend config

## Description

Adding onBeforeEventSend config

## Related Issue

https://jira.corp.adobe.com/browse/CORE-37564

## Types of changes

## Checklist:

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -84,6 +84,10 @@
                   "type": "string",
                   "enum": ["web", "device", "environment", "placeContext"]
                 }
+              },
+              "onBeforeEventSend": {
+                "type": "string",
+                "pattern": "^%[^%]+%$"                  
               }
             },
             "required": [
--- a/src/view/components/wrappedField.jsx
+++ b/src/view/components/wrappedField.jsx
@@ -44,16 +44,20 @@ const getNestedValue = (obj, path) => {
 const noop = () => {};
 
 export class DecoratedInput extends React.Component {
-  openDataElementSelector = replace => {
+  openDataElementSelector = () => {
     const {
-      field: { onChange, value = "", name }
+      field: { onChange, value = "", name },
+      supportDataElement
     } = this.props;
 
     // Whenever we're dealing with a data element token, we add it to whatever the existing value
     // is. If we're not dealing with a token, we replace the value entirely. This is just due
     // to how we want the UX to flow.
     window.extensionBridge.openDataElementSelector().then(dataElement => {
-      const newValue = replace ? dataElement : `${value}${dataElement}`;
+      const newValue =
+        supportDataElement === "replace"
+          ? dataElement
+          : `${value}${dataElement}`;
       onChange(newValue, { target: { value: newValue, name } });
     });
   };
@@ -127,9 +131,7 @@ export class DecoratedInput extends React.Component {
           <Button
             variant="tool"
             icon={<Data />}
-            onClick={() =>
-              this.openDataElementSelector(supportDataElement.replace)
-            }
+            onClick={this.openDataElementSelector}
           />
         ) : null}
       </ValidationWrapper>
@@ -146,7 +148,7 @@ DecoratedInput.propTypes = {
   // eslint-disable-next-line react/forbid-prop-types
   form: PropTypes.object.isRequired,
   children: PropTypes.node,
-  supportDataElement: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
+  supportDataElement: PropTypes.string,
   errorTooltipPlacement: PropTypes.oneOf(["top", "right", "bottom", "left"]),
   onChange: PropTypes.func,
   onBlur: PropTypes.func
--- a/src/view/components/wrappedField.jsx
+++ b/src/view/components/wrappedField.jsx
@@ -44,7 +44,7 @@ const getNestedValue = (obj, path) => {
 const noop = () => {};
 
 export class DecoratedInput extends React.Component {
-  openDataElementSelector = () => {
+  openDataElementSelector = replace => {
     const {
       field: { onChange, value = "", name }
     } = this.props;
@@ -53,7 +53,7 @@ export class DecoratedInput extends React.Component {
     // is. If we're not dealing with a token, we replace the value entirely. This is just due
     // to how we want the UX to flow.
     window.extensionBridge.openDataElementSelector().then(dataElement => {
-      const newValue = `${value}${dataElement}`;
+      const newValue = replace ? dataElement : `${value}${dataElement}`;
       onChange(newValue, { target: { value: newValue, name } });
     });
   };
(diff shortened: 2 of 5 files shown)


src/view/configuration/configuration.jsx
```diff
@@ -528,7 +537,23 @@ const Configuration = () => {
                           </div> */}
 
                           <h3>Context</h3>
-
+                          <div className="u-gapTop">
+                            <InfoTipLayout tip="If you want to add, remove, or modify fields from the event globally, you can configure an onBeforeEventSend callback.  This callback will be called everytime and event is sent.  This callback passes an object with an `xdm` fields.  Modify the xdm object to change the data that is sent in the event.">
+                              <FieldLabel
+                                labelFor="onBeforeEventSendField"
+                                label="Set this to configure a callback that will be called for every event."
```
Aaronius: It's a bit odd for a label to be telling the customer what to do. It should probably describe the feature. This is a tricky one though. Maybe something like "Callback function for modifying data before each event is sent to the server" (no period to be consistent with the other labels)

src/view/configuration/configuration.jsx
```diff
@@ -528,7 +537,23 @@ const Configuration = () => {
                           </div> */}
 
                           <h3>Context</h3>
-
+                          <div className="u-gapTop">
+                            <InfoTipLayout tip="If you want to add, remove, or modify fields from the event globally, you can configure an `onBeforeEventSend` callback. This callback will be called everytime an event is sent. This callback passes an object with a `xdm` fields. Modify the `xdm` object to change the data that is sent in the event.">
+                              <FieldLabel
```
jonsnyder: a `xdm` fields -> an xdm field

src/view/configuration/configuration.jsx
```diff
@@ -528,7 +537,23 @@ const Configuration = () => {
...
+                              />
+                            </InfoTipLayout>
+                            <div>
+                              <WrappedField
+                                id="onBeforeEventSendField"
+                                name={`instances.${index}.onBeforeEventSend`}
+                                component={Textfield}
+                                componentClassName="u-fieldLong"
+                                supportDataElement
+                              />
```
jonsnyder: @Aaronius What do you think about disabling this textbox?  Do people ever type in the name of the data element?  Disabling the textbox would show that the only thing you can do is click the data element button.

jonsnyder: Having the button is weird though because it will make two data elements concat-ed together if you click it twice.

Aaronius: Yes, customers do sometimes type the name of the data element. I've seen this type of thing come up in Slack conversations with external customers. I would support adding an option to wrappedField that would replace the existing value with the selected data element rather than append, though. I wouldn't want to use it everywhere (sometimes users do something like `%product name% - %product model%`, but in the case of this field (and some others in this extension), it would make sense to just replace the value. Maybe instead of `supportDataElement` create two separate options: `supportAppendDataElement` and `supportReplaceDataElement` or something like that.

src/view/components/wrappedField.jsx
```diff
@@ -44,7 +44,7 @@ const getNestedValue = (obj, path) => {
 const noop = () => {};
 
 export class DecoratedInput extends React.Component {
-  openDataElementSelector = () => {
+  openDataElementSelector = replace => {
     const {
       field: { onChange, value = "", name }
     } = this.props;
```
Aaronius: You can pull `replace` directly off props here instead of passing as it argument if you want.

extension.json
```diff
@@ -77,6 +77,10 @@
                   "type": "string",
                   "enum": ["web", "device", "environment", "placeContext"]
                 }
+              },
+              "onBeforeEventSend": {
+                "type": "string",
+                "pattern": "^%[^%]+%$"
```
jfkhoury: Maybe a comment here would help why that pattern is necessary.

src/view/configuration/configuration.jsx
```diff
@@ -153,6 +155,9 @@ const validateDuplicateValue = (createError, instances, key, message) => {
 const idSyncContainerIdValidationMessage =
   "Please specify a non-negative integer or data element for the container ID.";
 
+const onBeforeEventSendValidationMessage =
+  "Please specify a data element for the onBeforeEventSend callback.";
```
jfkhoury: Did we agree to wrap the vars in ` ` ? Like `onBeforeEventSend? cc: @Aaronius

extension.json
Aaronius: He can't comment here unfortunately since it's a JSON file. :(

src/view/configuration/configuration.jsx
Aaronius: I'm glad you bring this up, because in this case, `onBeforeEventSend` means nothing to the user, since we don't use that label for the form field. The label @jileeshs chose to use for the form field is "Callback function for modifying data before each event is sent to the server". So either we say "Please specify a data element for the callback function for modifying data before each event is sent to the server" or we change the field's label to something simpler and use that in the validation message instead.

The validation errors point directly at the field with the problem, so having the name of the field doesn't provide much value and is one more place we need to keep updated when we modify field labels. At this point, I think it would be good to just drop the field labels from all the validation messages entirely. For example change, "Please specify a non-negative integer or data element for the container ID." to "Please specify a non-negative integer or data element." and change "Please specify a data element for the onBeforeEventSend callback." to "Please specify a data element." Let's do it across the whole extension though.

Unless someone disagrees?

src/view/configuration/configuration.jsx
```diff
@@ -168,6 +170,9 @@ const validateDuplicateValue = (createError, instances, key, message) => {
 const idSyncContainerIdValidationMessage =
   "Please specify a non-negative integer or data element for the container ID.";
 
+const onBeforeEventSendValidationMessage =
+  "Please specify a data element.";
```
Aaronius: Will you please do the same for the all the other validation messages so that we're consistent in our approach? You can do it in a different PR if you'd like.

src/view/configuration/configuration.jsx
```diff
@@ -345,7 +354,7 @@ const Configuration = () => {
                                 name={`instances.${index}.name`}
                                 component={Textfield}
                                 componentClassName="u-fieldLong"
-                                supportDataElement
+                                supportDataElement="replace"
```
Aaronius: Do you mind taking a look at the other `supportDataElement` usage within the extension and seeing if they should use `supportDataElement="replace"` as well? For example, it would make sense for all the usage in the Send Event action to be `replace` as well. To manually test the actions, read this first: https://github.com/adobe/reactor-extension-alloy#manually-testing-views-using-the-local-sandbox

Feel free to do this in a separate PR if you'd like

PR #47, opened 2019-12-03
CORE-37247 Adding click data collection configuration options

liljenback:
## Description

Adding click data collection configuration options

## Related Issue

CORE-37247

## Motivation and Context

Configure everything

## Types of changes

## Checklist:

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -75,7 +75,8 @@
                 "type": "boolean"
               },
               "downloadLinkQualifier": {
-                "type": "string"
+                "type": "string",
+                "minLength": 1
               },
               "context": {
                 "type": "array",
--- a/extension.json
+++ b/extension.json
@@ -71,6 +71,12 @@
                 "type": "string",
                 "minLength": 1
               },
+              "clickCollectionEnabled": {
+                "type": "boolean"
+              },
+              "downloadLinkQualifier": {
+                "type": "string"
+              },
               "context": {
                 "type": "array",
                 "items": {
--- a/extension.json
+++ b/extension.json
@@ -3,7 +3,7 @@
   "name": "adobe-alloy",
   "iconPath": "resources/images/icon.svg",
   "platform": "web",
-  "version": "0.0.1",
+  "version": "0.0.3",
   "description": "The Adobe Experience Platform Web SDK allows for streaming data into the platform, syncing identities, personalizing content, and more.",
   "author": {
     "name": "Adobe"
--- a/src/view/configuration/configuration.jsx
+++ b/src/view/configuration/configuration.jsx
@@ -58,7 +58,8 @@ const getInstanceDefaults = initInfo => ({
   context: contextOptions,
   idMigrationEnabled: true,
   clickCollectionEnabled: true,
-  downloadLinkQualifier: "\\.(exe|zip|wav|mp3|mov|mpg|avi|wmv|pdf|doc|docx|xls|xlsx|ppt|pptx)$"
+  downloadLinkQualifier:
+    "\\.(exe|zip|wav|mp3|mov|mpg|avi|wmv|pdf|doc|docx|xls|xlsx|ppt|pptx)$"
 });
 
 const createDefaultInstance = initInfo =>
@@ -205,11 +206,9 @@ const validationSchema = object()
           }
           return validator;
         }),
-        downloadLinkQualifier: string()
-        .test({
+        downloadLinkQualifier: string().test({
           name: "invalidDownloadLinkQualifier",
-          message:
-            "Please provide a valid regular expression.",
+          message: "Please provide a valid regular expression.",
           test(value) {
             try {
               return new RegExp(value) !== null;
@@ -218,7 +217,7 @@ const validationSchema = object()
             }
           }
         })
-    })
+      })
     )
   })
   // TestCafe doesn't allow this to be an arrow function because of
--- a/src/view/configuration/configuration.jsx
+++ b/src/view/configuration/configuration.jsx
@@ -46,7 +46,7 @@ const getInstanceDefaults = initInfo => ({
   name: "alloy",
   configId: "",
   orgId: initInfo.company.orgId,
-  edgeDomain: "beta.adobedc.net",
+  edgeDomain: "edge.adobedc.net",
   errorsEnabled: true,
   optInEnabled: false,
   idSyncEnabled: true,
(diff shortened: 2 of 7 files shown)


src/view/configuration/configuration.jsx
```diff
@@ -527,8 +543,50 @@ const Configuration = () => {
                             </div>
                           </div> */}
 
-                          <h3>Context</h3>
+                          <h3>Data Collection</h3>
 
+                          <div className="u-gapTop">
+                            <InfoTipLayout tip="Indicates whether data associated with clicks on navigational links, download links, or personalized content should be automatically collected or not.">
```
Aaronius: The "or not" at the end here is redundant since it's implied by the "whether".

src/view/configuration/configuration.jsx
```diff
@@ -527,8 +543,50 @@ const Configuration = () => {
...
+                          </div>
+                          {values.instances[index].clickCollectionEnabled ? (
+                            <div>
+                              <InfoTipLayout tip="Regular expression that qualifies a link URL as a download link.">
+                                <FieldLabel
+                                  labelFor="downloadLinkQualifier"
+                                  label="Download Link Qualifier"
+                                />
+                              </InfoTipLayout>
+                              <div>
```
Aaronius: Let's add the class `FieldSubset` on this container, so that the disclosed download link qualifier field is inset a bit. This is what we did for the context checkboxes and was an agreed upon pattern with Laurel from UX.

extension.json
```diff
@@ -71,6 +71,12 @@
                 "type": "string",
                 "minLength": 1
               },
+              "clickCollectionEnabled": {
+                "type": "boolean"
+              },
+              "downloadLinkQualifier": {
+                "type": "string"
```
Aaronius: What do you think about validating that it's a `minLength` of `1` and adding the respective validation to https://github.com/adobe/reactor-extension-alloy/pull/47/files#diff-9e7d2c5e07f687fc721afa2ef6683586R209? An empty string seems like a mistake that a user would want to know about. I don't think they would want every link to be considered a download, for example.

test/functional/configuration/configuration.spec.js
```diff
@@ -118,7 +129,8 @@ test("initializes form fields with full settings", async t => {
             urlDestinationsEnabled: true,
             cookieDestinationsEnabled: false,
             prehidingStyle: "#container { display: none }",
-            context: ["device", "placeContext"]
+            context: ["device", "placeContext"],
+            clickCollectionEnabled: false
```
Aaronius: It would probably make sense to add `downloadLinkQualifier` with some custom regex to the second instance object, just so you can assert that the field gets populated with a custom value.

src/view/configuration/configuration.jsx
```diff
@@ -108,7 +111,9 @@ const getSettings = ({ values, initInfo }) => {
         "urlDestinationsEnabled",
         "cookieDestinationsEnabled",
         "prehidingStyle",
-        "idMigrationEnabled"
+        "idMigrationEnabled",
+        "clickCollectionEnabled",
+        "downloadLinkQualifier"
```
Aaronius: I think we should only add `downloadLinkQualifier` to the settings object if `clickCollectionEnabled` is `true`, since it's useless data otherwise. This is the approach we've taken for `idSyncContainerId` and `context` below.

You'll also need to validate the `downloadLinkQualifier` only when click collection is enabled. You'll see where this is currently a problem if you provide an invalid regular expression pattern, then uncheck the "Enable click data collection" checkbox, then hit save. It doesn't pass validation but there's no indication of what's wrong and, technically, nothing _is_ wrong.

test/functional/configuration/configuration.spec.js
```diff
@@ -407,6 +435,16 @@ test("restores default edge domain value when restore button is clicked", async
...
+test("restores default download link qualifier when restore button is clicked", async t => {
+  await extensionViewController.init(t, defaultInitInfo);
+  await instances[0].downloadLinkQualifierField.typeText(t, ".");
+  await instances[0].downloadLinkQualifierRestoreButton.click(t);
+  await instances[0].downloadLinkQualifierField.expectValue(
+    t,
+    defaultDownloadLinkQualifier
+  );
+});
+
```
Aaronius: You'll need to add to the "returns full valid settings" test. For the first Alloy instance, act like the user deselected the checkbox and for the other Alloy instance, act like the user provided a custom regex.

You'll also need to add a test to check that an error is shown for an invalid regular expression pattern (e.g., `[`), but only when click collection is enabled.

Aaronius:
Nice work so far!

Aaronius:
One thing I failed to mention is that Launch provides a regular expression tester that you could offer to users. You can see it in action if you go add a Value Comparison condition (offered by the Core extension) and choose "Matches Regex" for the operator:

image

If you click the "Test" link it will pop up the regular expression tester:

image

I think it would useful to offer it to our users. The "Test" link in the Core extension isn't Spectrum-compliant, so you'll probably want to just use a quiet button instead (like our "Restore to default" buttons). I'm not sure if "Test" is a sufficient label either. Maybe "Test Regex" or something like that would be better?

Here are the docs on how to leverage the regular expression tester: https://developer.adobelaunch.com/extensions/reference/views/#windowextensionbridgeopenregextesteroptions-object--promise

liljenback:
@Aaronius , I thought about having a regex tester popup just like that one. I didn't know one existed in Launch. I will see if I can include it.

test/functional/configuration/configuration.spec.js
```diff
@@ -393,6 +424,14 @@ test("shows error for duplicate IMS org ID", async t => {
...
 });
 
+test("shows error for invalid download link qualifier", async t => {
+  await extensionViewController.init(t, defaultInitInfo);
+  await instances[0].downloadLinkQualifierField.clear(t);
+  await instances[0].downloadLinkQualifierField.typeText(t, "[");
+  await extensionViewController.expectIsNotValid(t);
+  await instances[0].downloadLinkQualifierField.expectError(t);
+});
+
```
Aaronius: I still think it would be good to test that downloadLinkQualifierField doesn't get flagged as invalid if clickCollectionEnabledField is unchecked. Do you agree?

src/view/configuration/configuration.jsx
```diff
@@ -527,8 +558,75 @@ const Configuration = () => {
...
+                              <div>
+                                <WrappedField
+                                  id="downloadLinkQualifierField"
+                                  name={`instances.${index}.downloadLinkQualifier`}
+                                  component={Textfield}
+                                  componentClassName="u-fieldLong"
+                                />
+                                <Button
+                                  id="downloadLinkQualifierTestButton"
+                                  className="RightButton"
```
Aaronius: There are some theory discussions to be had here, but I'd probably recommend using `u-gapLeft` instead of `RightButton` for this case.

PR #61, opened 2020-03-11
CORE-25307 Fixed warnings for extension configuration view.

Aaronius:
## Description
I previously upgraded react and react-spectrum and started seeing the following warnings when running the extension configuration view:

```
configuration.8bb28e70.js:20967 Warning: Cannot update a component from inside the function body of a different component.
    in Formik (created by ExtensionView)
    in ErrorBoundary (created by ExtensionView)
    in ExtensionView (created by Configuration)
    in Configuration
    in div (created by Provider)
    in Provider
```

and another warning from React-Spectrum about how the quiet _variant_ of Button was deprecated.

I fixed both warnings.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-42299

## Types of changes

## Checklist:


src/view/configuration/configuration.jsx
```diff
@@ -230,457 +232,456 @@ const validationSchema = object()
     );
   });
 
-const Configuration = () => {
-  const [selectedAccordionIndex, setSelectedAccordionIndex] = useState();
-  const [isFirstExtensionViewRender, setIsFirstExtensionViewRender] = useState(
```
Aaronius: The problem was that we were using these react hooks from the `render` function that's fed into `ExtensionView`, which React didn't like (but didn't actually break anything). I split the render code into its own component and moved to hooks into that component. I also leveraged the `useNewlyValidatedFormSubmission` hook I created for the XDM Object data element view, which simplified some of the code as well.

src/view/configuration/configuration.jsx
```diff
@@ -230,457 +232,456 @@ const validationSchema = object()
...
-                                  const instanceDefaults = getInstanceDefaults(
-                                    initInfo
-                                  );
-                                  setFieldValue(
-                                    `instances.${index}.orgId`,
-                                    instanceDefaults.orgId
-                                  );
-                                }}
-                                quiet
-                                variant="quiet"
```
Aaronius: This is what the react-spectrum warning was about. Removing the `variant` attribute fixed the issue. The `quiet` attribute we already had provides appropriate "quiet" styling.

PR #62, opened 2020-03-17
Reduces view size for XDM Object view. Removes source maps.

Aaronius:
## Description
Reduces view size for XDM Object view. Removes source maps.

I didn't trim as much as I was hoping. It's still at 2.1 MB, which is rather large, whereas the other views are ~500-600 KB. I upgraded to Ant v4, which introduced better tree-shaking, but apparently it's still not great and for the tree component it also introduced features we may not need (e.g., virtual rendering). Anyway, I tried to use parcel-plugin-bundle-visualizer to see if I might be able to figure out how to cut it down further, but the report it generated was empty. I spent about a half hour trying to get it to work and cut my losses.

I then tried to cut size by enabling scope hoisting in Parcel, but that was throwing errors. It turns out it was throwing errors due to something in Ant's tree component. There's a [github issue on this](https://github.com/parcel-bundler/parcel/issues/3487) and someone [mentioned that it was fixed in Parcel 2](https://github.com/parcel-bundler/parcel/issues/3487#issuecomment-600095099), but that requires a bigger effort to upgrade Parcel to a not-yet-GA version.

I then noticed that Ant v4 deprecated the way we're rendering children for the Tree component, but there's nothing mentioned about it in documentation (I only saw the warning in the console). The alternative is to use the `treeData` prop on the Tree itself and let the Tree render the nodes, but they fail to say how one would go about providing a custom tree node title renderer in this case.

Anyway, I decided to move forward with Ant v4. You'll see a deprecation warning in the console, which I feel is fine ignoring for now until we receive more information on how to move away from it or react-spectrum releases a more appropriate tree component.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-42538

## Motivation and Context
Build sizes are large.

## Types of changes

## Checklist:

### DIFF
--- a/scripts/helpers/build.js
+++ b/scripts/helpers/build.js
@@ -35,7 +35,8 @@ module.exports = (options = {}) => {
       outDir: viewOutDir,
       watch,
       // HMR seems to be broken: https://github.com/parcel-bundler/parcel/issues/2894
-      hmr: false
+      hmr: false,
+      sourceMaps: false
     });
 
     bundler.on("bundled", () => {
--- a/test/functional/dataElements/xdmObject/helpers/xdmTree.js
+++ b/test/functional/dataElements/xdmObject/helpers/xdmTree.js
@@ -37,7 +37,7 @@ export default {
     const node = await getNode(title);
     await t.click(
       node
-        .parent("li")
+        .parent(".ant-tree-treenode")
         .nth(0)
         .find(".ant-tree-switcher")
     );
--- a/package-lock.json
+++ b/package-lock.json
@@ -173,37 +173,32 @@
         "tinycolor2": "^1.4.1"
       }
     },
-    "@ant-design/create-react-context": {
-      "version": "0.2.5",
-      "resolved": "https://registry.npmjs.org/@ant-design/create-react-context/-/create-react-context-0.2.5.tgz",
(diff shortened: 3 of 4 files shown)


test/functional/dataElements/xdmObject/helpers/xdmTree.js
```diff
@@ -37,7 +37,7 @@ export default {
     const node = await getNode(title);
     await t.click(
       node
-        .parent("li")
```
Aaronius: The tree component doesn't use `li`s anymore.

scripts/helpers/build.js
```diff
@@ -35,7 +35,8 @@ module.exports = (options = {}) => {
       outDir: viewOutDir,
       watch,
       // HMR seems to be broken: https://github.com/parcel-bundler/parcel/issues/2894
-      hmr: false
+      hmr: false,
```
liljenback, line 38: So we shouldn't just use an older version until it is resolved? Or maybe we can add a ticket to change it back?

Aaronius: I feel like there's more value in staying on the latest version of Parcel than saving HMR. I've created a JIRA ticket for switching HMR back on once it's fixed in Parcel: https://jira.corp.adobe.com/browse/CORE-44089

Aaronius:
Thanks for the review!

PR #65, opened 2020-03-19
Convert to using test IDs in tests

Aaronius:
## Description
Convert to using test IDs in tests. This is so we have a consistent way of referencing DOM elements that is less fragile than other options.

In full disclosure, for a proper e2e test, it's usually recommended to reference elements by what the user sees (the label, field placeholder, etc.). This is because, in an end-to-end test, one "end" is the user, and the user looks for the button that says "Add", for example, rather than looking for the button with some attribute or class name (that's what developers do). This is a relevant explanation: https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

However, there are downsides to this. It's often more costly to maintain and in many cases it's very difficult to achieve this for certain elements, which means we have inconsistencies in how we select elements. As we discussed this as a team, it was decided we would use `data-test-id` attributes for selecting DOM elements.

I have three commits: One where I switch everything over to `data-test-id` attributes. One where I add built-in-support for test IDs inside our `spectrum` helper module, which allows us to just pass in a test ID (or a `Selector` if necessary). The last commit removes `data-test-id` attributes from production builds since they're only needed for e2e tests.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-42303

## Motivation and Context
Consistency and simplicity.

## Types of changes

## Checklist:

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,9 @@
 {
   "plugins": ["@babel/plugin-proposal-class-properties"],
-  "presets": [
-    ["@babel/preset-env"]
-  ]
+  "presets": [["@babel/preset-env"]],
+  "env": {
+    "production": {
+      "plugins": ["babel-plugin-jsx-remove-data-test-id"]
+    }
+  }
 }
--- a/src/view/components/checkboxList.jsx
+++ b/src/view/components/checkboxList.jsx
@@ -38,20 +38,16 @@ class CheckboxList extends React.Component {
   render() {
     const { options, value, className } = this.props;
     const listItems = options.map(option => {
-      let optionValue;
-      let optionLabel;
-
-      if (typeof option === "string") {
-        optionValue = option;
-        optionLabel = option;
-      } else {
-        optionValue = option.value;
-        optionLabel = option.label;
-      }
+      const {
+        value: optionValue,
+        label: optionLabel,
+        testId: optionTestId
+      } = option;
 
       return (
         <li key={optionValue}>
           <Checkbox
+            data-test-id={optionTestId}
             value={optionValue}
             checked={value && value.includes(optionValue)}
             onChange={this.onChange}
--- a/test/functional/actions/sendEvent.spec.js
+++ b/test/functional/actions/sendEvent.spec.js
@@ -13,20 +13,15 @@ governing permissions and limitations under the License.
 import createExtensionViewController from "../helpers/createExtensionViewController";
 import spectrum from "../helpers/spectrum";
 import testInstanceNameOptions from "../helpers/testInstanceNameOptions";
-import { createTestIdSelector } from "../helpers/dataTestIdSelectors";
 
 const extensionViewController = createExtensionViewController(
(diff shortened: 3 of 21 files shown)


src/view/components/checkboxList.jsx
```diff
@@ -38,20 +38,16 @@ class CheckboxList extends React.Component {
   render() {
     const { options, value, className } = this.props;
     const listItems = options.map(option => {
-      let optionValue;
-      let optionLabel;
-
-      if (typeof option === "string") {
```
Aaronius: This was so we could just pass an array of values for the `options` props like:

`options={["cookie", "cake", "ice cream"]}`

We're not using this and I don't know that it buys us much, so I took it out for simplicity.

.babelrc
```diff
@@ -1,6 +1,9 @@
 {
   "plugins": ["@babel/plugin-proposal-class-properties"],
-  "presets": [
-    ["@babel/preset-env"]
-  ]
+  "presets": [["@babel/preset-env"]],
+  "env": {
+    "production": {
+      "plugins": ["babel-plugin-jsx-remove-data-test-id"]
```
jonsnyder: nice!

test/functional/actions/sendEvent.spec.js
```diff
@@ -10,19 +10,18 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
 const extensionViewController = createExtensionViewController(
   "actions/sendEvent.html"
 );
-const instanceNameField = spectrum.select(Selector("[name=instanceName]"));
-const viewStartField = spectrum.checkbox(Selector("[name=viewStart]"));
-const xdmField = spectrum.textfield(Selector("[name=xdm]"));
-const typeField = spectrum.textfield(Selector("[name=type]"));
-const mergeIdField = spectrum.textfield(Selector("[name=mergeId]"));
+const instanceNameField = spectrum.select("instanceNameField");
```
jonsnyder: Nice! This is much cleaner now that you just have to pass in a string.

test/functional/actions/setCustomerIds.spec.js
```diff
@@ -31,27 +30,18 @@ const mockExtensionSettings = {
   ]
 };
 
-const instanceNameField = spectrum.select(Selector("[name=instanceName]"));
-const addCustomerIdButton = spectrum.button(Selector("#addCustomerId"));
```
jonsnyder: Probably don't need the id property on the add customer id button anymore.

Aaronius: Where did you see one? I believe it was removed: https://github.com/adobe/reactor-extension-alloy/pull/65/files#diff-73b166646fdf3d8b15c143f08ada5d96L40

jonsnyder: Sorry I was mistaken.

Aaronius:
Yeah, it's a good idea. I'll do that before merging.

PR #66, opened 2020-04-03
CORE-43397 Added event merge ID caching and action for resetting.

Aaronius:
## Description
There's some history to this one but I'll try to keep it short. Previously, the Event Merge ID data element would produce a new merge ID every time the data element was used. This is not good because the whole purpose of an event merge ID is that you can pass the same value when sending different events.

The "Storage Duration" option in Launch would not change this behavior, since Launch's data element storage is a fallback cache instead of an up-front cache. In other words, Launche's data element storage would only be consulted if the Event Merge ID data element returned `null` or `undefined`, which would never happen. Even if the Launch user could use the "Storage Duration" option in Launch, it would probably be insufficient since the only options are None, Page View, Session, and Visitor and the website owner may want to reset the event merge ID at more specific points in time.

For this reason, we had to build in a storage mechanism inside the Event Merge ID data element itself, and then provide a Reset Event Merge ID action that allows the customer to reset the event merge ID at a specific point in time.

Ideally, we would have been able to use the data element name as the cache key. But, Turbine does not provide the data element library module with the name of the data element for which the library module is producing an event merge ID. So, we had two options: (1) change Turbine so that it passes the data element name to the data element library module or (2) generate an ID during data element creation and save it to the settings object for the data element, then use the generated ID as the cache key. Number one would take more time and there are some things we would need to sort out about the best way to go about passing the data element name into the data element library module. I'm still open to it, but due to time constraints we decided to go with option 2. It would probably take me a day to modify Turbine to provide the data element name to the library module and update documentation, extension tools and such, though I would legitimately spend more time than that anguishing about _how_ we provide the name to the library module due to some Turbine API baggage that makes consistency difficult, then the change would need to get deployed to Launch production which would probably take a few days.

For anyone currently using the Event Merge ID data element, they'll need to go back into each of their Event Merge ID data elements and re-save the data element in order for the storage to begin working for that data element.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-43397
https://jira.corp.adobe.com/browse/CORE-43398

## Types of changes

## Checklist:

- [ ] **I don't think we have extension-specific documentation?** I have updated the documentation accordingly.


src/view/dataElements/xdmObject.styl
```diff
@@ -15,7 +15,3 @@ governing permissions and limitations under the License.
 .XdmObject-treeContainer {
   width: 300px;
 }
-
-.XdmObject-description {
-  font-size: 18px;
-}
```
Aaronius, line 21: This is unrelated to the rest of the PR. I noticed it should be in noSelectedNodeView.styl.

src/view/dataElements/eventMergeId.jsx
```diff
@@ -0,0 +1,77 @@
...
+              <FieldLabel labelFor="instanceNameField" label="Instance" />
+              <div>
+                <WrappedField
+                  data-test-id="instanceNameField"
+                  id="instanceNameField"
+                  name="instanceName"
+                  component={Select}
+                  componentClassName="u-fieldLong"
+                  options={getInstanceOptions(initInfo)}
+                />
```
Aaronius, line 67: Technically, from within the data element library module, it's possible for us to just use the first instance we find to generate an event merge ID, since a random ID will be generated regardless of which instance we use. We still ask for which instance the customer wants to use to generate the event merge ID though because there will be log messages in the console that say which instance the `createEventMergeId` command is being run within and it would be weird if a customer used one instance 99% of the time and then sees a console log saying that a different instance was used to generate an event merge ID.

src/view/actions/resetEventMergeId.jsx
```diff
@@ -12,47 +12,56 @@ governing permissions and limitations under the License.
...
-                id="instanceNameField"
-                name="instanceName"
-                component={Select}
+                data-test-id="eventMergeIdField"
+                id="eventMergeIdField"
+                name="eventMergeId"
+                component={Textfield}
                 componentClassName="u-fieldLong"
-                options={getInstanceOptions(initInfo)}
+                supportDataElement="replace"
```
Aaronius, line 64: Notice we do _not_ ask for an instance on this action. That's because there's really no need to, since the cache is not specific to any particular instance.

src/lib/actions/resetEventMergeId/createResetEventMergeId.js
```diff
@@ -10,17 +10,8 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-// disablePageReloads is not a publicized feature, but it sure helps speed up tests.
-// https://github.com/DevExpress/testcafe/issues/1770
-fixture("ECID and Event Merge ID View").disablePageReloads.page(
-  "http://localhost:3000/viewSandbox.html"
-);
-
-testInstanceNameOnlyView(extensionViewController);
+module.exports = eventMergeIdCache => {
+  return settings => {
+    eventMergeIdCache.clearByEventMergeId(settings.eventMergeId);
```
jonsnyder, line 15: Can you add a comment explaining why we clear here instead of reset?

// The next time the data element is accessed, a new eventMergeId will be created.

src/view/dataElements/eventMergeId.jsx
jonsnyder: One way to get rid of the instance selector is to just generate the id inside the launch code, and not call the alloy function.  We could log the creation in the satellite logger.  The instance selector makes sense for sending an event, but not really for the eventMergeId.

Aaronius: Yeah, that would work. I wouldn't want to have duplicate uuidv4 code (one in the extension and one in alloy), so I'll probably want to make the reactor-specific Alloy build use the uuidv4 from the extension. Launch extensions don't support importing npm packages yet either, so that might make things tricky. Let me dig deeper and see what I find.

Aaronius:
Alright, I've updated it so that the user does not need to select an instance on the event merge ID data element. If you recall, in the reactor-specific build we expose a configuration option called `reactorRegisterCreateEventMergeId`. This is so that the instance can provide the extension a function that will create an event merge ID synchronously. The alloy instance [calls `reactorRegisterCreateEventMergeId`](https://github.com/adobe/alloy/blob/c82445ac2cfa72052667bceb560a1ad2f3bd5843/src/components/EventMerge/index.js#L21) like this:

```
config.reactorRegisterCreateEventMergeId(uuid);
```

In other words, every instance calls `reactorRegisterCreateEventMergeId` with the same function: the `uuid` function, which is not instance-specific in any way.

Because of this, I've restructured the "instance manager" in the extension a little so that the instance manager exposes a `createEventMergeId` method that the Event Merge ID data element can call _without_ needing an instance name. Calling `instanceManager.createEventMergeId`, in essence, is just calling the `uuid` function provided by the last-created Alloy instance. I would recommend taking a look at the changes to `lib/instanceManager/createInstanceManager.js` first and then the rest of the changes will probably make more sense.

src/lib/instanceManager/createInstanceManager.js
```diff
@@ -30,36 +32,38 @@ module.exports = ({ turbine, window, runAlloy, orgId }) => {
       // provides a backdoor to perform certain operations
       // synchronously, because Reactor requires that data
       // elements be resolved synchronously for now.
-      reactorRegisterGetEcid(getEcid) {
-        accessor.getEcid = getEcid;
-      },
```
Aaronius, line 35: Notice I removed the getEcid stuff until we figure out how we're actually going to handle ECIDs (and it sounds like the command name will be changing too).

src/view/dataElements/eventMergeId.styl
```diff
@@ -10,7 +10,8 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 */
 
-const createEcid = require("./createEcid");
-const instanceManager = require("../../instanceManager/index");
+@import "../global";
 
-module.exports = createEcid({ instanceManager, turbine });
+.EventMergeId-description {
+  font-size: 18px;
+}
```
Aaronius, line 17: This diff is misleading since this file has nothing to do with `ecid/index.js`. I removed the getEcid stuff until we figure out how we're actually going to handle ECIDs (and it sounds like the command name will be changing too). The ECID stuff we did have was not actually being used (it wasn't referenced from `extension.json`).

Aaronius:
Me too! Thanks for pushing me on it.

PR #68, opened 2020-04-07
CORE-37918 Sync SDK debugging with Launch debugging.

Aaronius:
## Description
This syncs the SDK debugging with Launch debugging, such that when a Launch customer toggles Launch debugging, the SDK debugging will also be toggled. ~~This is dependent upon new Turbine functionality introduced in https://github.com/adobe/reactor-turbine/pull/103, and therefore should not be merged until the Turbine PR is merged and deployed to production.~~ **Update: https://github.com/adobe/reactor-turbine/pull/103 has been published to production. We can merge this PR once it has received two reviews.**

## Related Issue
https://jira.corp.adobe.com/browse/CORE-37918

## Motivation and Context
There's some history behind this one. For a while, the reactor-specific build of Alloy would use the logger provided by Turbine for logging messages. We would configure Alloy instances to always have `debugEnabled` set to `true`, so that Alloy would always be logging messages through the logger provided by Turbine. Turbine would then only print the messages to the console when Launch debugging was enabled by the user. Just to emphasize the point: this required us to always have `debugEnabled` set to `true`.

Later, we started doing more than just logging when `debugEnabled` was set to true. Specifically, we would flip on a synchronous validation flag in our requests to the edge, which would force Platform to validate data synchronously against the configured XDM schema and return any validation errors in the response. Because of this, we didn't want to have `debugEnabled` set to `true` all the time on the Alloy instances because synchronous validation was only intended to be used under debugging scenarios (it's expensive). So, we switched Alloy instances _back_ to not having `debugEnabled` set to `true` and not using the logger provided by Turbine. This meant that customers would have to manually run `alloy("debug", { enabled: true })` in the console if they wanted to enable Alloy debugging, which kind of sucks.

Since then, we changed how we do synchronous validation. We no longer flip a synchronous validation flag in our requests to the edge. Instead, synchronous validation will be enabled through a query string parameter that will get appended to requests through the Adobe Experience Cloud Debugger browser extension.

Because of all of the above, debugging in Alloy now only toggles whether we log messages. I don't think we're doing anything beyond that when Alloy debugging is enabled. So...one option is we go back to setting `debugEnabled` to `true` always when we configure Alloy instances and use the logger from Turbine from within Alloy all the time. BUT, I think we may run into similar scenarios in the future where `debugEnabled` doesn't mean that _only_ logging is enabled and we'll find ourselves in this awkward situation again. So, I think it's best if we actually toggle Alloy's debugging when Launch's debugging is toggled. There are some other, smaller benefits to this approach too as well. For example, in some cases we skip some computation, like serializing certain objects for log messages, if we know that debugging is disabled.

Note that syncing Alloy's debugging with Launch's debugging does not preclude us from using the logger from Turbine. We could still use the logger from Turbine and only pump messages through it when Alloy's debugging is enabled. For better or worse, that would mean that any log messages from Alloy would be printed to the console with the 🚀 emoji and `[AEP Web SDK]` that Launch typically prepends, along with the message that Alloy logs. At the current moment, including in this PR, we are not using the logger from Turbine from within Alloy instances, but I'm not sold either way on what we should do here. I believe this is something we have flexibility with changing later if we want.

## Types of changes

## Checklist:

- [x] (Do we have extension documentation?) My change requires a change to the documentation.

### DIFF
--- a/src/lib/instanceManager/createInstanceManager.js
+++ b/src/lib/instanceManager/createInstanceManager.js
@@ -12,15 +12,19 @@ governing permissions and limitations under the License.
 
 module.exports = ({ turbine, window, runAlloy, orgId }) => {
   const accessorByInstanceName = {};
-  const { instances } = turbine.getExtensionSettings();
-  const names = instances.map(instance => instance.name);
+  const { instances: instancesSettings } = turbine.getExtensionSettings();
+  const instanceNames = instancesSettings.map(
+    instanceSettings => instanceSettings.name
+  );
 
-  runAlloy(names);
+  runAlloy(instanceNames);
 
-  instances.forEach(({ name, ...options }) => {
+  instancesSettings.forEach(({ name, ...options }) => {
+    const instance = window[name];
     const accessor = {};
-    window[name]("configure", {
+    instance("configure", {
       ...options,
+      debugEnabled: turbine.debugEnabled,
       orgId: options.orgId || orgId,
       // The Alloy build we're using for this extension
       // provides a backdoor to perform certain operations
@@ -33,8 +37,11 @@ module.exports = ({ turbine, window, runAlloy, orgId }) => {
         accessor.createEventMergeId = createEventMergeId;
       }
     });
-    accessor.instance = window[name];
+    accessor.instance = instance;
     accessorByInstanceName[name] = accessor;
+    turbine.onDebugChanged(enabled => {
+      instance("debug", { enabled });
+    });
   });
 
   return {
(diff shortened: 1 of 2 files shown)


Aaronius:
Thanks for the review!

Aaronius:
This PR was set up to merge from the `toggle-debug` branch to the `inject-turbine` branch. The `inject-turbine` had already been merged though, so merging this PR wasn't effective at updating master. I've since manually merged the `toggle-debug` branch to master.

PR #69, opened 2020-04-20
CORE-41768 Changed onBeforeEventSend field to code editor.

Aaronius:
## Description
Changed onBeforeEventSend field to code editor. Previously, it was a textfield that accepted a data element.

Notice I added a "placeholder" prop to EditorButton which we're now using for both `prehidingStyle` and `onBeforeEventSend`. If the user does not modify the placeholder, no code will be saved. I'm curious to know how you feel about this.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-41768

## Motivation and Context
It's simper for a user to provide code "inline" instead of having to go create a custom code data element.

## Types of changes

## Checklist:

- [x] (probably?) My change requires a change to the documentation.

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -73,7 +73,7 @@
               },
               "onBeforeEventSend": {
                 "type": "string",
-                "pattern": "^%[^%]+%$"
+                "minLength": 1
               }
             },
             "required": ["configId", "name"],
@@ -83,7 +83,14 @@
       },
       "required": ["instances"],
       "additionalProperties": false
-    }
+    },
+    "transforms": [
+      {
+        "type": "function",
+        "propertyPath": "instances[].onBeforeEventSend",
+        "parameters": ["content"]
+      }
+    ]
   },
   "actions": [
     {
--- a/src/view/components/editorButton.jsx
+++ b/src/view/components/editorButton.jsx
@@ -23,17 +23,32 @@ import PropTypes from "prop-types";
  */
 class EditorButton extends React.Component {
   onClick = () => {
-    const { onChange, value, language } = this.props;
+    const { onChange, value, language, placeholder } = this.props;
 
     const options = {
-      code: value
+      code: value || placeholder || ""
     };
 
     if (language) {
       options.language = language;
     }
 
-    window.extensionBridge.openCodeEditor(options).then(onChange);
+    window.extensionBridge.openCodeEditor(options).then(updatedCode => {
+      // A bug exists in the Launch UI where the promise is resolved
+      // with undefined if the user hits Cancel. In this case, the promise
+      // should never be resolved or rejected.
+      // https://jira.corp.adobe.com/browse/DTM-14454
+      if (updatedCode === undefined) {
+        return;
+      }
+
+      // If the user never changed placeholder code, don't save the placeholder code.
+      if (placeholder && updatedCode === placeholder) {
+        updatedCode = "";
+      }
+
+      onChange(updatedCode);
+    });
   };
 
   render() {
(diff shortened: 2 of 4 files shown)


extension.json
```diff
@@ -83,7 +83,14 @@
       },
       "required": ["instances"],
       "additionalProperties": false
-    }
+    },
+    "transforms": [
+      {
+        "type": "function",
+        "propertyPath": "instances[].onBeforeEventSend",
```
Aaronius, line 90: Support for array notation on `propertyPath` is the functionality we were waiting for from Launch.

src/view/configuration/configuration.jsx
```diff
@@ -580,9 +578,13 @@ const Configuration = ({ formikProps, initInfo }) => {
                                   pattern: currentPattern
                                 })
                                 .then(newPattern => {
-                                  values.instances[
-                                    index
-                                  ].downloadLinkQualifier = newPattern;
```
Aaronius, line 585: We shouldn't ever attempt to modify Formik state directly. That's what the `setFieldValue` call on the next line is for (which also ensures that views get re-rendered, etc.).

extension.json
jfkhoury: Hey @Aaronius, is there a public doc we can link to?

src/view/configuration/configuration.jsx
```diff
@@ -580,9 +578,13 @@ const Configuration = ({ formikProps, initInfo }) => {
...
+                                  // A bug exists in the Launch UI where the promise is resolved
+                                  // with undefined if the user hits Cancel. Instead, the promise
+                                  // should have never been resolved or rejected.
+                                  // https://jira.corp.adobe.com/browse/DTM-14454
```
Aaronius, line 584: I tried creating an issue in JIRA for removing this code once DTM-14454 is resolved, but JIRA is struggling. I'll try later.

extension.json
Aaronius: Yes, there's this: https://developer.adobelaunch.com/extensions/reference/extension-manifest/#transforming-settings-objects

But comments aren't supported in JSON. Would you like to see a comment in configuration.jsx somewhere?

jfkhoury: Thank you! Only if you think it's useful.

src/view/components/editorButton.jsx
```diff
@@ -23,17 +23,32 @@ import PropTypes from "prop-types";
...
+        return;
+      }
+
+      // If the user never changed placeholder code, don't save the placeholder code.
+      if (placeholder && updatedCode === placeholder) {
+        updatedCode = "";
+      }
+
+      onChange(updatedCode);
+    });
```
Aaronius, line 51: This is where the code editor placeholder magic happens.

src/view/configuration/configuration.jsx
Aaronius: Issue has been added: https://jira.corp.adobe.com/browse/CORE-44498

src/view/configuration/configuration.jsx
```diff
@@ -535,7 +535,7 @@ const Configuration = ({ formikProps, initInfo }) => {
                           component={EditorButton}
                           language="javascript"
                           placeholder={
-                            '// Modify content.xdm as necessary. For example:\n// content.xdm.product = "shirt";'
+                            '// Modify content.xdm as necessary. There is no need to wrap the code in a function\n// or return a value. For example:\n// content.xdm.product = "shirt";'
```
jfkhoury, line 538: Nice! 👍

jonsnyder: I know this is just an example, but content.xdm.product isn't a valid xdm field.  We could go with
```content.xdm.web.webPageDetails.name = "Checkout";```

Aaronius: Done. Thanks!

PR #70, opened 2020-04-21
CORE-43465 Return promise from actions. Throw error instead of logging.

Aaronius:
## Description
**Tests are failing until https://github.com/adobe/reactor-extension-alloy/pull/72 is merged.**

This change is forward-looking for rule component sequencing support that will be coming in Launch. The change is that we return promises from the actions. Also, rather than logging an error when we can't find the Alloy instance by name, we now throw an error. This is probably more appropriate, since Launch will catch the error, log it, and prevent subsequent actions within the rule from executing (which, according to Launch users, would typically be the expected outcome when an action fails).

## Related Issue
https://jira.corp.adobe.com/browse/CORE-43465

## Motivation and Context
Support rule component sequencing.

## Types of changes

## Checklist:

- [ ] **Tests are failing until https://github.com/adobe/reactor-extension-alloy/pull/72 is merged.** All tests pass and I've made any necessary test changes.

### DIFF
--- a/test/unit/lib/dataElements/eventMergeId/createEventMergeId.spec.js
+++ b/test/unit/lib/dataElements/eventMergeId/createEventMergeId.spec.js
@@ -16,12 +16,8 @@ describe("Event Merge ID", () => {
   let eventMergeIdCache;
   let instanceManager;
   let dataElement;
-  let turbine;
 
   beforeEach(() => {
-    turbine = {
-      logger: jasmine.createSpyObj("logger", ["error"])
-    };
     instanceManager = jasmine.createSpyObj("instanceManager", {
       createEventMergeId: "randomEventMergeId"
     });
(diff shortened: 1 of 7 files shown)


test/unit/lib/dataElements/eventMergeId/createEventMergeId.spec.js
```diff
@@ -31,8 +27,7 @@ describe("Event Merge ID", () => {
     ]);
     dataElement = createEventMergeId({
       instanceManager,
-      eventMergeIdCache,
-      turbine
```
Aaronius, line 35: `turbine` wasn't actually being used within this data element.

Aaronius:
@jonsnyder yeah, thanks for bringing that up. To answer your question, Launch does not run the second action if the first gets rejected (this was driven by unanimous vote by Launch users, though I think errors in alloy are more pervasive that what they're currently working with due to how we throw an error when the user has opted out). I'll bring these cases up with Justin and Ben Robison just to make sure they're aware and see how they would like to handle it.

PR #75, opened 2020-04-29
CORE-44969 Changing setCustomerIds command name and options.

Aaronius:
## Description
Changes to match https://github.com/adobe/alloy/pull/467.

Old API:

```
alloy("setCustomerIds", {
  // IDs here
});
```

New API:

```
alloy("syncIdentity", {
  identities: {
    // IDs here
  }
});
```

## Related Issue
https://jira.corp.adobe.com/browse/CORE-44969

## Motivation and Context
Driven by product requirements. We're nesting identities to give more flexibility to add other options later.

## Types of changes

## Checklist:


extension.json
```diff
@@ -215,16 +210,17 @@
...
           }
         },
-        "required": ["instanceName", "customerIds"],
+        "required": ["instanceName", "identities"],
         "additionalProperties": false
       },
-      "libPath": "dist/lib/actions/setCustomerIds/index.js",
-      "viewPath": "actions/setCustomerIds.html"
+      "libPath": "dist/lib/actions/syncIdentity/index.js",
+      "viewPath": "actions/syncIdentity.html"
```
Aaronius, line 223: Previously, `customerIds` was an array and each customer ID object had a `namespace` property on it. When the Set Customer Ids action was run, we would transform the array into a object where keys were the namespaces and the values were the customer ID objects.

I presume the reason why `customerIds` in settings was an array is because that's how the customer IDs were managed within the view state (specifically, inside formik). However, I think it's most appropriate to not model the settings object according to view state, but instead what will be sent through Alloy, for a couple reasons:

1. It requires less transformation within code deployed to customer websites, since the data will already be in a map instead of an array.
2. While we can validate inside the extension view that there are no two identities with the same namespace, you can't validate that for Launch API users if we were to use the schema we previously had. If, however, our schema dictates that `identities` an object where the keys are the namespaces, however, the structure itself enforces that API users _can't_ get into a situation where they have multiple identities with the same namespace.

src/view/actions/syncIdentity.jsx
```diff
@@ -16,34 +16,56 @@ import { array, object, string } from "yup";
...
+  });
+};
+
+const identitiesArrayToMap = identitiesArray => {
+  return identitiesArray.reduce((identityMap, identity) => {
+    const { namespace, ...otherProperties } = identity;
+    identityMap[identity.namespace] = otherProperties;
+    return identityMap;
+  }, {});
+};
```
Aaronius, line 38: This is the magic where we translate between our application state structure and our settings structure.

src/view/components/identityWrapper.jsx
```diff
@@ -0,0 +1,151 @@
...
+    </React.Fragment>
+  );
+}
+
+IdentityWrapper.propTypes = {
+  values: PropTypes.object,
+  initInfo: PropTypes.object
+};
+
+export default IdentityWrapper;
```
Aaronius, line 151: It's unfortunate that git shows this as a whole new file. Really, it was renamed from customerIdWrapper.jsx and modified. Other than the "customer ID" rename, I also simplified some of the code because the checks like:

```
{((Array.isArray(values.customerIds) &&	
                values.customerIds.length) ||	
                null)
```

were unnecessary since `values.customerIds` (now `values.identities`) is always defined and has at least one item.

Aaronius:
@jileeshs you may want to review this as well since you're probably most familiar with the area.

PR #77, opened 2020-04-30
Types support

Aaronius:
## Description

I'm so sorry this is a big pull request. 🙏 Most of it was moving code around and renaming some things. Here are some details:

This adds support in the XDM Object data element for number, integer, and boolean values without having to use data elements (though data elements can still be used).

As I added more types than what we already supported (object, array, and string), it became apparent that our helpers were getting too large and complex because the logic for all the various types were getting mixed into the same modules. So, I split out type-specific logic into separate, smaller helpers that the "master" helpers delegate to. Logic that is common to all types still resides in the "master" helpers.

I also renamed `wholeValue` to just `value`. I'll try to explain why. It's important to understand that each node has a concept of a "population strategy" and either a `whole` or `parts` population strategy is used. A `whole` population strategy means the user has elected to provide the value (either via a data element or a constant) of the node directly. All types of nodes can use the `whole` population strategy. A `parts` population strategy, on the other hand, means the user has elected to compose the node by providing values for its individual parts--in other words, individual items (in the case of an array node) or individual properties (in the case of an object node). Only objects and arrays support the `parts` population strategy. As more types were added (number, integer, boolean) that only support the `whole` population strategy, the term `wholeValue` to represent the directly-provided value started to have less significance (and a little confusing for such node types) and just `value` seemed sufficient. So, I renamed `wholeValue` to just `value` within the form state. In the case of objects and array nodes, if the user chooses the `whole` population strategy, the value of `value` is saved to the settings object for the node. If the user chooses the `parts` population strategy, the object or array that's saved to the settings object is composed using the values of its parts (or in other words, its descendant nodes).

I also added e2e tests for the new types and fixed an issue where our e2e tests would fail if we made any changes to the schema within Platform (because we were asserting a hard-coded version). More e2e tests are warranted to check for things like validation, but this should be sufficient for this week's release.

@alshaw I fully expect we'll have conflicts between your PR and my PR, so when you get to that point, let's get together and work through them.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-43416

## Motivation and Context
We have users wanting to provide constant values for number, integer, and boolean fields.

## Types of changes

## Checklist:

- [ ] **All of the XDM Object data element will need to be documented later as part of https://jira.corp.adobe.com/browse/CORE-42500** I have updated the documentation accordingly.


src/view/dataElements/xdmObject/components/booleanEdit.jsx
```diff
@@ -0,0 +1,161 @@
...
+      <RadioGroup
+        id="inputMethodField"
+        selectedValue={inputMethod}
+        onChange={newInputMethod => {
+          onChange("");
+          setInputType(newInputMethod);
+        }}
+      >
+        <Radio label="Data Element" value={DATA_ELEMENT} />
+        <Radio className="u-gapLeft2x" label="Constant" value={CONSTANT} />
```
jfkhoury: Are we going to rename this to `Enter Value`?

Aaronius: You were looking at one of the older commits. The latest commit has "Select Value".

Aaronius:
I demoed it in Sprint Demo, but I'd be happy to demo it again or review the code structure a little more. I'll try to remember to bring it up in a tech checkin. Thanks for reviewing. I know it sucks to review big PRs.

PR #80, opened 2020-05-05
Adds schema selection feature to xdmObject

garyr:
## Description

Adds a schema selection feature to xdmObject data object page. Allows an extension user to choose a schema from a list prior to editing the schema

## Types of changes

## Checklist:


src/view/dataElements/xdmObject/helpers/findSchemaMetaById.js
```diff
@@ -0,0 +1,22 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export default ({ $id, schemasMeta }) => {
+  let foundSchemaMeta;
+  for (let i = 0; i < schemasMeta.length; i += 1) {
```
jfkhoury, line 15: Can't we use `schemasMeta.find(meta => meta.$id === $id)` ?

garyr: Yes, that's a better implementation. I will change it

src/view/dataElements/xdmObject.jsx
```diff
@@ -79,45 +113,102 @@ const XdmObject = ({ initInfo, formikProps, schema }) => {
...
+
+        return (
+          <XdmObject
+            {...options}
+            schemasMeta={schemasMeta}
+            selectedSchemaMeta={selectedSchemaMeta}
+            setSelectedSchemaMeta={onSchemaMetaSelected}
+          />
+        );
+      }}
```
Aaronius, line 211: I haven't thought this through much, but it looks like it's about time to start breaking this module down into smaller modules. This isn't something I would recommend doing before GA, but maybe after.

src/view/dataElements/xdmObject.jsx
```diff
@@ -23,53 +25,85 @@ import NodeEdit from "./xdmObject/components/nodeEdit";
...
+          id="schemaField"
+          data-test-id="schemaField"
+          className="u-widthAuto u-gapBottom"
+          options={schemaOptions}
+          value={selectedSchemaMeta.$id}
+          onChange={schemaMetaId => {
+            setSelectedNodeId(undefined);
+            setSelectedSchemaMeta({
+              $id: schemaMetaId
+            });
```
Aaronius, line 78: Although you can get away with this currently, it introduces inconsistency in what is considered a "schema meta". Is it an object with `$id`? Is it an object with both `$id` and `title`? To be consistent and predictable, I recommend looking up the actual schema meta object using the ID before setting state:

```
const matchingSchemaMeta = schemaMetas.find(schemaMeta => schemaMeta.$id === schemaMetaId);
setSelectedSchemaMeta(matchingSchemaMeta);
```

src/view/dataElements/xdmObject.jsx
```diff
@@ -79,45 +113,102 @@ const XdmObject = ({ initInfo, formikProps, schema }) => {
...
+  const [selectedSchemaMeta, setSelectedSchemaMeta] = useState();
 
   return (
     <ExtensionView
       getInitialValues={({ initInfo }) => {
-        return fetchSchema({
-          edgeConfigId: initInfo.extensionSettings.instances[0].edgeConfigId,
+        // TODO: If we've already loaded the schemas, then don't load the
+        // schemas again. Just call getInitialValues with the newly
+        // selected schema and return the result.
```
Aaronius, line 130: I'm not sure I understand this TODO. I may have even written it. 😆😢I think it may no longer be pertinent?

src/view/dataElements/xdmObject.jsx
```diff
@@ -79,45 +113,102 @@ const XdmObject = ({ initInfo, formikProps, schema }) => {
...
+        return fetchSchemaList({
```
Aaronius, line 131: Consider renaming this `fetchSchemasMeta` just to be consistent with what the promise is being resolved with.

src/view/dataElements/xdmObject.jsx
```diff
@@ -79,45 +113,102 @@ const XdmObject = ({ initInfo, formikProps, schema }) => {
...
+            if (!_schemasMeta.length) {
+              return {};
+            }
 
-          const initialValues = getInitialValues({
-            settings: initInfo.settings,
-            schema: fetchedSchema
-          });
+            // TODO: don't initialize form state if the schema doesn't match
+            let existingSchema;
```
Aaronius, line 143: Maybe rename to `existingSchemaMeta` or, more explicitly, `schemaMetaMatchingPreviousSave`. The main point is if it's a schema meta instead of a full schema, then it's probably important to keep the "meta" part in the name.

src/view/dataElements/xdmObject.jsx
```diff
@@ -79,45 +113,102 @@ const XdmObject = ({ initInfo, formikProps, schema }) => {
...
+              existingSchema = findSchemaMetaById({
+                $id: initInfo.settings.schema.id,
+                schemasMeta: _schemasMeta
+              });
+            }
+
+            let schemaMeta = _schemasMeta[0];
+            if (existingSchema) {
+              schemaMeta = existingSchema;
+            }
```
Aaronius, line 154: Consider something like this:

```
let schemaMeta = _schemasMeta[0];

if (initInfo.settings && initInfo.settings.schema) {
  schemaMeta = findSchemaMetaById({
    $id: initInfo.settings.schema.id,
    schemasMeta: _schemasMeta
  });
}
```

src/view/dataElements/xdmObject.jsx
```diff
@@ -79,45 +113,102 @@ const XdmObject = ({ initInfo, formikProps, schema }) => {
...
         };
       }}
       validate={validate}
-      render={props => <XdmObject {...props} schema={schema} />}
+      render={options => {
+        const onSchemaMetaSelected = _schemaMeta => {
+          const schemaMeta = findSchemaMetaById({
+            $id: _schemaMeta.$id,
+            schemasMeta
+          });
```
Aaronius, line 188: This is related to my other comment. Keep the translation from the selected schema ID to the matching schema meta as close to the Select as possible. That way, the argument passed for `_schemaMeta` on this `onSchemaMetaSelected` function is actually a schema meta rather than just an object with an ID that's sort of like a schema meta but not really.

src/view/dataElements/xdmObject.jsx
```diff
@@ -79,45 +113,102 @@ const XdmObject = ({ initInfo, formikProps, schema }) => {
...
+            id: selectedSchemaMeta.$id,
+            version: selectedSchemaMeta.version
           },
           data: getValueFromFormState({ formStateNode: values }) || {}
         };
       }}
       validate={validate}
-      render={props => <XdmObject {...props} schema={schema} />}
+      render={options => {
+        const onSchemaMetaSelected = _schemaMeta => {
```
Aaronius, line 184: It might be worth adding a TODO for handling race conditions when the user selects a schema then selects another schema before the first schema is done loading. We either need to cancel the previous request or disable the Schema dropdown while the schema is loading.

src/view/dataElements/xdmObject.jsx
```diff
@@ -122,35 +120,27 @@ const XdmExtensionView = () => {
...
-        return fetchSchemaList({
+        return fetchSchemasMeta({
           orgId: initInfo.company.orgId,
           imsAccess: initInfo.tokens.imsAccess
         })
           .then(_schemasMeta => {
             setSchemasMeta(_schemasMeta);
 
-            if (!_schemasMeta.length) {
+            if (!_schemasMeta || !_schemasMeta.length) {
```
Aaronius, line 134: Hey @garyr, when do you suspect `_schemasMeta` will be falsey? I'm wondering if this is a case where the promise should be rejected instead of resolved.

garyr: This might just be an issue better handled elsewhere, but when I have a bad access token, and the fetchSchemasMeta call fails, I believe the variable is null and doesn't have a length property. This causes an unknown length property error, instead of the better error message

garyr: I updated with better handling of the 401 conditions

PR #86, opened 2020-05-19
Rebase the `alloy-latest-release` after tagging Alloy.

jfkhoury:
`alloy-latest-release` is the branch we will start using to run the Prod E2E tests. We were using `master` in the past but that was causing issues because we might have updated testing specs in master that don't match the PROD released Alloy. @Aaronius

## Types of changes

## Checklist:


Aaronius:
I mentioned this over Slack, but rather than maintaining a git branch, I'd rather our CI looked up the most recent tag in git and used that. It sounds like that will require a bit of custom scripting though so if we need to maintain a branch in the meantime then that's fine.

jfkhoury:
> I mentioned this over Slack, but rather than maintaining a git branch, I'd rather our CI looked up the most recent tag in git and used that. It sounds like that will require a bit of custom scripting though so if we need to maintain a branch in the meantime then that's fine.

I agree 100%! I created a ticket to automate that process (Unless there's a way in Travis) https://jira.corp.adobe.com/browse/CORE-46402

PR #92, opened 2020-07-08
Travis requires the travis.yml to be in the root of the project.

shammowla:
The reactor extension is setup with [Travis](https://travis-ci.com/github/adobe/reactor-extension-alloy/settings) but the travis.yml file is in a sub directory. 

The cron and build is most likely not triggering because the travis.yml isn't in the root. 

https://stackoverflow.com/questions/31768540/how-to-use-travis-without-a-travis-yml-file-in-the-root-folder-of-a-repository#:~:text=Travis%20CI%20uses%20.,it%2C%20or%20configure%20it%20elsewhere.


jonsnyder:
I don't think react-spectrum wants you to publish their code to an open source repo.

PR #98, opened 2020-07-29
Adds platform sandbox UI

garyr:
## Description
Adds the ability to choose a platform sandbox and schemas by sandbox within the data elements page. Upon choosing a sandbox and schema, those settings will persist in the extension settings

## Motivation and Context
This is a requested feature to ease customer adoption of Alloy

## Types of changes

## Checklist:


garyr:
Please don't merge until @alshaw completes her CSS changes. Also I am going to try to add some functional tests for this feature...

garyr:
I updated this PR to include a bug fix where some schemas were getting 404 responses. This was caused by schema loads not including the sandbox name in the request

src/view/dataElements/xdmObject/helpers/fetchSandboxes.js
```diff
@@ -0,0 +1,51 @@
...
+    type: "production",
+    isDefault: true,
+    region: null,
+    state: "active"
+  };
+
+  return fetch(
+    `${platform.getHost()}/data/foundation/sandbox-management/sandboxes`,
+    {
+      headers: {
```
jonsnyder, line 33: you can simplify this: "headers: baseRequestHeaders"

src/view/dataElements/xdmObject/helpers/fetchSandboxes.js
```diff
@@ -0,0 +1,51 @@
...
+        ...baseRequestHeaders
+      }
+    }
+  )
+    .then(response => {
+      if (response.status === SANDBOX_PERMISSION_ERROR) {
+        return {
+          sandboxes: [DEFAULT_SANDBOX],
```
jonsnyder, line 41: Can you add a comment here explaining why we use a default sandbox for a 403 error?

src/view/dataElements/xdmObject/helpers/fetchSchemasMeta.js
```diff
@@ -13,17 +13,31 @@ governing permissions and limitations under the License.
...
+  } else {
+    headers["x-sandbox-name"] = platform.getDefaultSandbox();
+  }
+
   // TODO: paginate this response using on responseBody._page.count or responseBody._links.next
   return fetch(
-    `${platform.getHost()}/data/foundation/schemaregistry/tenant/schemas?orderby=title`,
+    `${platform.getHost()}/data/foundation/schemaregistry/tenant/schemas?orderby=title&property=meta:extends==${metaExtends}`,
     {
       headers: {
```
jonsnyder, line 39: simplify to "{ headers }"

src/view/dataElements/xdmObject/helpers/getInitialFormState.js
```diff
@@ -64,6 +64,7 @@ const generateNodeId = () => {
  * and is made available for convenient access so it doesn't have to
  * be repeatedly computed on every render.
  *
+ * @param {Object} sandboxMeta The sandbox meta object.
```
jonsnyder, line 67: It's a little bit unclear to me why this is called "meta".  It looks like you were following a naming pattern that Aaron had setup here. Is it just "Details about the selected sandbox"?

src/view/dataElements/xdmObject.jsx
```diff
@@ -57,101 +69,210 @@ const XdmObject = ({
...
           imsAccess: initInfo.tokens.imsAccess
         })
-          .then(_schemasMeta => {
-            setSchemasMeta(_schemasMeta);
+          .then(_sandboxesMeta => {
+            let sandboxName = null;
+            // if sandboxes are returned from the api call
+            if (_sandboxesMeta.sandboxes && _sandboxesMeta.sandboxes.length) {
+              setSandboxesMeta(_sandboxesMeta);
+              // default to the first item in the list
```
jonsnyder, line 217: Should we default to the sandbox that has "isDefault: true"?

src/view/dataElements/xdmObject.jsx
```diff
@@ -164,38 +285,87 @@ const XdmExtensionView = () => {
       }}
       getSettings={({ values }) => {
         return {
+          sandbox: {
+            name: selectedSandboxMeta ? selectedSandboxMeta.name : null
+          },
```
Aaronius, line 290: Hey @garyr, it doesn't look like these changes are reflected in the schema defined in `extension.json`: https://github.com/adobe/reactor-extension-alloy/blob/master/extension.json#L258-L272

src/view/dataElements/xdmObject.jsx
```diff
@@ -57,101 +69,210 @@ const XdmObject = ({
...
+              sandboxName
+            })
+              .then(schema => {
+                setSchemaStatus(STATUS_LOADED || null);
+
+                return schema;
+              })
+              .catch(() => {
+                setSchemaStatus(STATUS_ERROR || null);
+              });
```
Aaronius, line 275: It looks like you can pull the `.then` and `.catch` up to the top-level of the promise chain rather than having it nested.

src/view/dataElements/xdmObject.jsx
```diff
@@ -57,101 +69,210 @@ const XdmObject = ({
...
+              imsAccess: initInfo.tokens.imsAccess,
+              sandboxName
+            });
+          })
+          .then(response => {
+            setSchemasMeta(response.results);
 
-            if (!_schemasMeta.length) {
+            if (!response.results.length) {
+              setSchemasMetaStatus(STATUS_ERROR || null);
```
Aaronius, line 242: What's the `|| null` for? `STATUS_ERROR` is a constant with a string value, so it doesn't seem like `|| null` would ever come into play.

src/view/dataElements/xdmObject/helpers/fetchSchema.js
```diff
@@ -13,17 +13,33 @@ governing permissions and limitations under the License.
...
   const baseRequestHeaders = getBaseRequestHeaders({ orgId, imsAccess });
   const schemaMajorVersion = schemaMeta.version.split(".")[0];
 
+  const headers = {
+    ...baseRequestHeaders,
+    // request a summary response with title , $id , meta:altId , and version attributes
+    Accept: "application/vnd.adobe.xdm-v2+json"
+  };
+
+  if (sandboxName && sandboxName !== "") {
```
Aaronius, line 30: These are redundant. `if (sandboxName) {` also covers `if (sandboxName !== "") {`

src/view/dataElements/xdmObject.jsx
Aaronius: BTW, you won't be able to make `sandbox` a new required field in the schema, because it would be considered backward-incompatible (https://developer.adobelaunch.com/extensions/reference/backwards-compatibility/). In other words, there are already settings objects in the database that don't have `sandbox` defined, so you can't change the schema to say that it's required. You may have already handled this (I haven't looked yet), but you'll need to make sure than you can load existing settings objects that don't contain the `sandbox` attribute into the view.

src/view/dataElements/xdmObject.jsx
Aaronius: If you do move the `.catch` up, just be aware that it will catch errors from `fetchSchema` but also any errors that happened higher up the chain, like errors from `fetchSchemasMeta`. You'll need to think through how you want to deal with the various errors that may occur anywhere in this promise chain.

src/view/dataElements/xdmObject.jsx
```diff
@@ -164,38 +285,87 @@ const XdmExtensionView = () => {
...
+          sandbox: {
+            name: selectedSandboxMeta ? selectedSandboxMeta.name : null
+          },
           schema: {
-            id: selectedSchemaMeta.$id,
-            version: selectedSchemaMeta.version
+            id:
+              selectedSchemaMeta && selectedSchemaMeta.$id
+                ? selectedSchemaMeta.$id
+                : null,
```
Aaronius, line 295: @garyr, do you mind describing when we get into `getSettings` and have a null `selectedSchemaMeta`? That seems like what this check is for anyway. It seems like we should make validation fail if the user hasn't selected a schema, right?

When the extension is running inside Launch, if the user hits Save, Launch calls `validate` on the extension view. If `validate` returns `false`, then Launch won't call `getSettings`. Launch only calls `getSettings` if `validate` returns `true`. So, in other words, if we make `validate` return false if no schema has been selected, then we shouldn't have to do this null check here in `getSettings`. Let me know if that doesn't make sense.

src/view/dataElements/xdmObject.jsx
garyr: Thanks. I will look into this...

src/view/dataElements/xdmObject.jsx
garyr: I have the null to get around this error that WebStorm is telling me without `|| null`:

```Argument type string is not assignable to parameter type ((prevState: undefined) => undefined) | undefined   Type string is not assignable to type (prevState: undefined) => undefined ```.  Can you suggest an alternative?

File: src/view/dataElements/xdmObject.jsx (hunk 8 above)
Comment (garyr) [reply src/view/dataElements/xdmObject.jsx:242]: I have the null to get around this error that WebStorm is telling me without `|| null`:

```
Argument type string is not assignable to parameter type ((prevState: undefined) => undefined) | undefined   Type string is not assignable to type (prevState: undefined) => undefined 
```

Can you suggest an alternative?

File: src/view/dataElements/xdmObject.jsx (hunk 10 above)
Comment (garyr) [reply src/view/dataElements/xdmObject.jsx:295]: Okay, I will look at addressing this

File: src/view/dataElements/xdmObject.jsx (hunk 5 above)
Comment (garyr) [reply src/view/dataElements/xdmObject.jsx:217]: Yes, I think so. I will make the change

File: src/view/dataElements/xdmObject.jsx (hunk 8 above)
Comment (Aaronius) [reply src/view/dataElements/xdmObject.jsx:242]: Hmm...that looks like error that comes from TypeScript checking. You might want to make sure TypeScript checking is disabled for the project. When I remove `|| null` in WebStorm, I don't see any error.

File: src/view/dataElements/xdmObject.jsx (hunk 7 above)
Comment (garyr) [reply src/view/dataElements/xdmObject.jsx]: Roger

File: src/view/dataElements/xdmObject.jsx (hunk 8 above)
Comment (garyr) [reply src/view/dataElements/xdmObject.jsx]: Okay, thanks. I removed the `|| null` instances

File: src/view/dataElements/xdmObject/helpers/fetchSchema.js (hunk 9 above)
Comment (garyr) [reply src/view/dataElements/xdmObject/helpers/fetchSchema.js]: ☑️

File: src/view/dataElements/xdmObject/helpers/getInitialFormState.js (hunk 4 above)
Comment (garyr) [reply src/view/dataElements/xdmObject/helpers/getInitialFormState.js:67]: Yes

File: src/view/dataElements/xdmObject/helpers/fetchSchemasMeta.js (hunk 3 above)
Comment (garyr) [reply src/view/dataElements/xdmObject/helpers/fetchSchemasMeta.js]: Done

File: src/view/dataElements/xdmObject/helpers/fetchSandboxes.js (hunk 2 above)
Comment (garyr) [reply src/view/dataElements/xdmObject/helpers/fetchSandboxes.js:42]: Done

File: src/view/dataElements/xdmObject/helpers/fetchSandboxes.js (hunk 1 above)
Comment (garyr) [reply src/view/dataElements/xdmObject/helpers/fetchSandboxes.js]: Done

File: src/view/dataElements/xdmObject.jsx [hunk 11]
```diff
@@ -164,38 +285,87 @@ const XdmExtensionView = () => {
...
+                ? selectedSchemaMeta.version
+                : null
           },
           data: getValueFromFormState({ formStateNode: values }) || {}
         };
       }}
       validate={validate}
       render={options => {
+        const onSandboxesMetaSelected = sandboxMeta => {
+          console.log(sandboxMeta);
```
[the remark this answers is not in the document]
garyr: Done

PR #104, opened 2020-08-07
Adds sandbox functional tests

garyr:
## Description
This change adds functional tests to cover most of the use cases for AEM sandbox manipulation in the xdm data elements view. This PR also changes the sandbox endpoint to return just the list the user has access to

## Motivation and Context
This change is needed to add automated testing coverage to recent features added to the extension

## Types of changes

## Checklist:

### DIFF
--- a/test/functional/dataElements/xdmObject.spec.js
+++ b/test/functional/dataElements/xdmObject.spec.js
@@ -86,9 +86,13 @@ const initializeExtensionView = async additionalInitInfo => {
 // disablePageReloads is not a publicized feature, but it sure helps speed up tests.
 // https://github.com/DevExpress/testcafe/issues/1770
 fixture("XDM Object View")
+  .beforeEach(async () => {
+    // adds a mocked sandboxes response
+    await t.addRequestHooks(platformMocks.sandboxes);
+  })
   .disablePageReloads.page("http://localhost:3000/viewSandbox.html")
   .meta("requiresAdobeIOIntegration", true)
-  .requestHooks(platformMocks.sandboxes);
+  .requestHooks(platformMocks.sandboxes)
 
 test("initializes form fields with individual object attribute values", async () => {
   await initializeExtensionView({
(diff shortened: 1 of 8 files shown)


test/functional/dataElements/xdmObject.spec.js
```diff
@@ -123,6 +109,54 @@ test("initializes form fields with individual object attribute values", async ()
...
+  // temporarily remove sandboxes mock
+  await t.removeRequestHooks(platformMocks.sandboxes);
+  // replace with unaurhotized mock
+  await t.addRequestHooks(platformMocks.sandboxesEmpty);
+  await initializeExtensionView();
+  await spectrum.select("sandboxField").expectDisabled();
+  await selectSchemaFromSchemasMeta();
+  await xdmTree.toggleExpansion("_alloyengineering");
+  // restore original sandboxes mock
+  await t.addRequestHooks(platformMocks.sandboxes);
```
Aaronius, line 122: Ideally, tests should be atomic, meaning if one test blows up it doesn't blow up other tests. If this test blows up (e.g., an error is thrown), this last line may not be called, causing subsequent tests to also fail. It might be better to do `await t.addRequestHooks(platformMocks.sandboxes);` inside a `beforeEach` rather than calling `.requestHooks()` on the fixture and having this line here.

test/functional/helpers/spectrum.js
```diff
@@ -253,12 +255,12 @@ Object.keys(componentWrappers).forEach(componentName => {
...
     return {
+      expectEnabled: createExpectEnabled(selector),
+      expectDisabled: createExpectDisabled(selector),
       ...componentWrapper.call(this, selector),
       selector,
       expectExists: createExpectExists(selector),
-      expectNotExists: createExpectNotExists(selector),
-      expectEnabled: createExpectEnabled(selector),
-      expectDisabled: createExpectDisabled(selector)
+      expectNotExists: createExpectNotExists(selector)
```
Aaronius, line 263: I like what you did here. I'd recommend moving `expectExists` and `expectNotExists` up as well so the individual component wrappers could override that if they wanted to, however unlikely.

test/functional/dataElements/xdmObject.spec.js
garyr: Yes, I see the problem now. I will make the change

garyr:
I implemented @Aaronius suggested changes

PR #108, opened 2020-09-01
CORE-51396: Adds environment aware platform api calls

garyr:
Users attempting to configure XDM Objects in non-production environments are unable to do so because of an authentication issue. The issue is caused by an api client that is unaware of the correct environment to use

## Description

This change modifies the platform api client to be environment aware

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/xdmObject/helpers/platform.js
+++ b/src/view/dataElements/xdmObject/helpers/platform.js
@@ -10,11 +10,42 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-const PROD = "https://platform.adobe.io";
+const PLATFORM_HOST_PROD = "https://platform.adobe.io";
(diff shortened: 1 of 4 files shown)


src/view/dataElements/xdmObject/helpers/platform.js
```diff
@@ -10,11 +10,35 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    if (!tokenParts[1]) {
+      return PLATFORM_HOST_PROD;
+    }
+
+    /**
+     * attempts to read `as` field from access token and use as environment reference
+     * NOTE: assumes production
+     */
+    const environment = imsAccess
+      ? JSON.parse(atob(tokenParts[1])).as
```
jonsnyder, line 35: What do you think about adding a try/catch for the syntax errors that could be thrown from JSON.parse?

garyr: Probably a good idea

PR #109, opened 2020-09-01
CORE-48075 - fetch namespaces from platform

ninaceban:
## Description

Now for the Identity Map data element, the user should enter the namespace code manually.
We should fetch namespaces from the platform and expose them in a dropdown. If there are no namespaces or the call failed we should allow the user to enter namespace code manually.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/identityMap.jsx
+++ b/src/view/dataElements/identityMap.jsx
@@ -22,13 +22,18 @@ import "./identityMap.styl";
 import { AMBIGUOUS } from "../utils/authenticatedState";
 import fetchNamespaces from "./identityMap/fetchNamespaces";
 
+const isNotECID = namespace => {
+  return namespace.code !== "ECID";
+};
 const identitiesMapToArray = identityMap => {
-  return Object.keys(identityMap).map(namespace => {
-    return {
-      namespace,
-      identifiers: identityMap[namespace]
-    };
-  });
+  return Object.keys(identityMap)
+    .sort((first, second) => first.localeCompare(second))
+    .map(namespace => {
+      return {
+        namespace,
+        identifiers: identityMap[namespace]
+      };
+    });
 };
 
 const identitiesArrayToMap = identitiesArray => {
@@ -49,10 +54,13 @@ const getInitialValues = ({ initInfo }) => {
     imsAccess: initInfo.tokens.imsAccess
   }).then(response => {
     if (response.length > 0) {
-      const namespaces = response.map(namespace => ({
-        value: namespace.code,
-        label: namespace.code
-      }));
+      const namespaces = response
+        .filter(isNotECID)
+        .map(namespace => ({
+          value: namespace.code,
+          label: namespace.code
+        }))
+        .sort((first, second) => first.value.localeCompare(second.value));
 
       return {
         identities,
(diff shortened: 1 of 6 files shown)


src/view/dataElements/identityMap/fetchNamespaces.js
```diff
@@ -0,0 +1,30 @@
...
+governing permissions and limitations under the License.
+*/
+
+import getBaseRequestHeaders from "../../utils/getBaseRequestHeaders";
+import platform from "../xdmObject/helpers/platform";
+
+export default ({ orgId, imsAccess }) => {
+  const headers = getBaseRequestHeaders({ orgId, imsAccess });
+
+  return fetch(`${platform.getHost()}/data/core/idnamespace/identities`, {
```
garyr, line 19: Hey @ninaceban, we will want to include the change that I just made in #108 to this

src/view/dataElements/identityMap/fetchNamespaces.js
```diff
@@ -0,0 +1,30 @@
...
+    headers
+  })
+    .then(response => {
+      if (!response.ok) {
+        return [];
+      }
+
+      return response.json();
+    })
+    .then(responseBody => responseBody);
```
jfkhoury, line 29: Should we add a `catch` and return an empty `[]` as well?

src/view/dataElements/identityMap.jsx
```diff
@@ -43,9 +44,29 @@ const getInitialValues = ({ initInfo }) => {
...
 
-  return {
-    identities
-  };
+  return fetchNamespaces({
+    orgId: initInfo.company.orgId,
+    imsAccess: initInfo.tokens.imsAccess
+  }).then(response => {
+    if (response.length > 0) {
+      const namespaces = response.map(namespace => {
```
jonsnyder, line 52: It's cleaner to remove the return statement:
const namespaces = response.map(namespace => ({
  value: namespace.code,
  label: namespace.code
}));

src/view/dataElements/identityMap/fetchNamespaces.js
ninaceban: thank you! I will update with latest

src/view/dataElements/identityMap.jsx
ninaceban: yes, thank you

ninaceban:
@jonsnyder and @Aaronius thank you for your suggestions, I have added more changes to the PR, could you please re-review it.
As per platform namespaces documentation, a namespace cannot be removed and a namespace code will never change once it was created.

jonsnyder (changes requested):
Here are a few UI suggestions after having run the sandbox for this branch:
* On initial load, expand the first accordion
* After a new identity is added, expand the accordion of the newly added identity
* Sort the namespaces in the drop down alphabetically
* Filter out ECID as a namespace from the drop down list
* Should we add some text to explain that ECID will be added automatically?
* When loading settings, sort the identities alphabetically by namespace
* Add an Infotip on the Authenticated State drop down
* Handle multiple identities with the same namespace. Either throw a validation error or generate the settings object correctly (currently, the second identity overwrites the first)
* Does there need to be any validation on the Primary checkbox? (I'm thinking that only one identifier should be primary)

jfkhoury:
@ninaceban Is this PR ready to be merged?

ninaceban:
@jonsnyder I have addressed most of your comments/suggestions except for those related to validation since `one primary ` validation if part of another PR.
Could you please take another look at the PR? Thank you

jonsnyder (approved):
Thank you. Looks great.

PR #112, opened 2020-09-09
Identity Map data element - validate multiple primary

ninaceban:
## Description

We should validate that user cannot send in `xdm.identityMap` more than one primary identifier.
## Types of changes

## Checklist:


src/view/dataElements/identityMap.jsx
```diff
@@ -74,6 +74,30 @@ const validateDuplicateValue = (
...
+  for (let i = 0; i < identities.length; i += 1) {
+    const { identifiers = [] } = identities[i];
+
+    for (let j = 0; j < identifiers.length; j += 1) {
+      if (hasPrimary(identifiers[j])) {
+        primaries += 1;
+      }
+
+      if (primaries > 1) {
+        return createError({
```
jonsnyder, line 91: Ideally, we would flag multiple primary checkboxes as invalid. For example, if three identifiers had primary checked, the 2nd and 3rd ones should be marked invalid. I think you can return an array of errors.

jonsnyder (approved):

src/view/dataElements/identityMap.jsx
ninaceban: I did it the way we check for duplicates, we show error for the first duplicate we found. If I change this one I think we should change the other validators as well. Either one works fine for me, I just want to be consistent.

jfkhoury (approved):

jfkhoury:
@ninaceban Does this PR need more work or can we merge as well?

PR #116, opened 2020-10-22
Adds ability to search and filter schema objects

garyr:
## Description
This feature allows for a user to search for and filter the list of schemas in the XDM Object view

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/xdmObject.jsx
+++ b/src/view/dataElements/xdmObject.jsx
@@ -158,15 +158,11 @@ const XdmObject = ({
               onBlur={() => {
                 if (
                   !selectedSchemaMeta ||
-                  selectedSchemaMeta.title !== schemaOptionsSearch
+                  !schemaOptionsSearch ||
+                  schemaOptionsSearch === ""
                 ) {
                   setSelectedNodeId(undefined);
-                  // try to find the selected schema
-                  setSelectedSchemaMeta(
-                    schemasMeta.find(
-                      schemaMeta => schemaMeta.name === schemaOptionsSearch
-                    )
-                  );
+                  setSelectedSchemaMeta(undefined);
                 }
               }}
               placeholder="Search for or select a schema"
@@ -363,10 +359,8 @@ const XdmExtensionView = () => {
         // TODO: address a suspected race condition where a previous selected item may return before
         //  a secondary selected item
         const onSchemaMetaSelected = schemaMeta => {
-          if (!schemaMeta) {
-            setSchemaStatus(STATUS_ERROR);
-          } else {
-            setSelectedSchemaMeta(schemaMeta);
+          setSelectedSchemaMeta(schemaMeta);
+          if (schemaMeta) {
             setSchemaOptionsSearch(schemaMeta.title);
             setSchemaStatus(STATUS_LOADING);
             fetchSchema({
@@ -386,6 +380,8 @@ const XdmExtensionView = () => {
               .catch(() => {
                 setSchemaStatus(STATUS_ERROR);
               });
+          } else {
+            setSchemaStatus(STATUS_LOADED);
           }
         };
 
--- a/src/view/dataElements/xdmObject.jsx
+++ b/src/view/dataElements/xdmObject.jsx
@@ -236,7 +236,6 @@ const XdmExtensionView = () => {
               setSelectedSandboxMeta(sandboxMeta);
             }
 
-            // return { "results": []};
             return fetchSchemasMeta({
               orgId: initInfo.company.orgId,
               imsAccess: initInfo.tokens.imsAccess,
--- a/src/view/dataElements/xdmObject.jsx
+++ b/src/view/dataElements/xdmObject.jsx
@@ -87,7 +87,6 @@ const XdmObject = ({
         search
       }).then(response => {
         if (!response.results.length) {
-          setSchemasMetaStatus(STATUS_ERROR);
           return;
         }
         setSchemasMeta(response.results);
@@ -358,18 +357,20 @@ const XdmExtensionView = () => {
             orgId: options.initInfo.company.orgId,
             imsAccess: options.initInfo.tokens.imsAccess,
             sandboxName: sandbox.name
-          }).then(response => {
-            if (!response.results.length) {
+          })
+            .then(response => {
+              if (!response.results.length) {
+                throw new Error("No schemas found");
+              }
+              setSchemasMeta(response.results);
+              setSchemaOptions(
+                getSchemaOptions({ schemasMeta: response.results })
+              );
+              setSelectedSchemaMeta(undefined);
+              setSchemasMetaStatus(STATUS_LOADED);
+            }).catch(() => {
               setSchemasMetaStatus(STATUS_ERROR);
-              return;
-            }
-            setSchemasMeta(response.results);
-            setSchemaOptions(
-              getSchemaOptions({ schemasMeta: response.results })
-            );
-            setSelectedSchemaMeta(undefined);
-            setSchemasMetaStatus(STATUS_LOADED);
-          });
+            });
         };
 
         // TODO: address a suspected race condition where a previous selected item may return before
--- a/src/view/dataElements/xdmObject.jsx
+++ b/src/view/dataElements/xdmObject.jsx
@@ -11,12 +11,14 @@ governing permissions and limitations under the License.
 */
 
 import "regenerator-runtime"; // needed for some of react-spectrum
-import React, { useState } from "react";
+import React, { useCallback, useState } from "react";
 import PropTypes from "prop-types";
 import Alert from "@react/react-spectrum/Alert";
 import ComboBox from "@react/react-spectrum/ComboBox";
 import Select from "@react/react-spectrum/Select";
 import FieldLabel from "@react/react-spectrum/FieldLabel";
+import Wait from "@react/react-spectrum/Wait";
+import debounce from "debounce";
 import NoSelectedNodeView from "./xdmObject/components/noSelectedNodeView";
 import ExtensionView from "../components/extensionView";
 import XdmTree from "./xdmObject/components/xdmTree";
@@ -28,6 +30,7 @@ import render from "../render";
 import fetchSandboxes from "./xdmObject/helpers/fetchSandboxes";
 import fetchSchema from "./xdmObject/helpers/fetchSchema";
 import fetchSchemasMeta from "./xdmObject/helpers/fetchSchemasMeta";
+import getSchemaOptions from "./xdmObject/helpers/getSchemaOptions";
 import "./xdmObject.styl";
 import InfoTipLayout from "../components/infoTipLayout";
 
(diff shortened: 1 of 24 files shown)


src/view/dataElements/xdmObject.jsx
```diff
@@ -115,22 +142,32 @@ const XdmObject = ({
...
-                  )
+                  schemasMeta.find(schemaMeta => schemaMeta.$id === item.value)
                 );
               }}
+              onMenuToggle={menuState => {
+                setShowComboMenu(menuState);
+              }}
+              placeholder="Search for or select a schema"
             />
+            <div className="u-gapTop u-gapLeft">
```
Aaronius, line 168: Rather than using `u-gapTop` here, I would recommend using `u-alignItemsCenter` on the div that contains this div. It will probably result in more accurate centering and is more robust if the size of the Wait or ComboBox changes.

src/view/dataElements/xdmObject.jsx
```diff
@@ -115,22 +142,32 @@ const XdmObject = ({
...
+              {schemaSearchInProgress === true && <Wait size="S" />}
```
Aaronius, line 169: `schemaSearchInProgress` is guaranteed to be a boolean so you can just do `schemaSearchInProgress && ...`

src/view/dataElements/xdmObject.jsx
```diff
@@ -207,111 +259,116 @@ const XdmExtensionView = () => {
...
         })
-          .then(_sandboxesMeta => {
-            let sandboxName = null;
+          .then(_sandboxes => {
+            let sandbox = { name: null };
             // if sandboxes are returned from the api call
-            if (_sandboxesMeta.sandboxes && _sandboxesMeta.sandboxes.length) {
-              setSandboxesMeta(_sandboxesMeta);
+            if (_sandboxes.sandboxes && _sandboxes.sandboxes.length) {
+              setSandboxes(_sandboxes);
```
Aaronius, line 266: The `sandboxes.sandboxes` thing can get a little confusing (not just here but in other parts of the code). What if we did `setSandboxes(_sandboxes.sandboxes)` and then had another state called `sandboxesEnabled` that you would set using `setSandboxesEnabled(!_sandboxes.disabled)`. I'd probably also rename `_sandboxes` here to `sandboxesResult` or do `.then({ sandboxes, disabled})` [here](https://github.com/adobe/reactor-extension-alloy/pull/116/files#diff-49c61fe83e69899661207037ed001ed1f94dacac52029c9b80e9c50000c882aeR262) and then you just avoid naming the object altogether. I hope that makes sense.

src/view/dataElements/xdmObject.jsx
```diff
@@ -61,19 +77,33 @@ const XdmObject = ({
...
+  const doSearch = useCallback(
+    debounce(({ sandboxName, search }) => {
+      setSchemaSearchInProgress(true);
+      return fetchSchemasMeta({
+        orgId: initInfo.company.orgId,
+        imsAccess: initInfo.tokens.imsAccess,
+        sandboxName,
+        search
+      }).then(response => {
+        if (!response.results.length) {
```
Aaronius, line 89: We need to take into consideration that a new request may be sent while a previous request is still active. In other words, if there's an active request and a subsequent request is made, we need to either cancel the active request (https://davidwalsh.name/cancel-fetch) or just ignore the response to the active request.

src/view/dataElements/xdmObject.jsx
```diff
@@ -115,22 +142,32 @@ const XdmObject = ({
               <FieldLabel labelFor="schemaField" label="Schema" />
             </InfoTipLayout>
           </div>
-          <div>
-            <Select
+          <div className="u-flex">
+            <ComboBox
```
Aaronius, line 146: In Slack you mentioned:

> There is a small issue when filtering the schema results and deleting the filter. After removing all the text search query, the combo box results do not update unless you click twice on the button to drop down the list.

I wanted to make sure you're aware that the bug might be a little more complex than this. Notice when I type "s" and then backspace and repeat the process:

searchconcurrency

Notice the results vary.

Aaronius (commented):
Looking good @garyr. I've left some feedback. I agree with your assessment that xdmObject.jsx is getting bloated and complex. It will take some thought to figure out how to best break it down and simplify it. I wouldn't let it hold up your PR from being merged though.

We do need to figure out the problem with the filter when clearing out the textfield. I think @jonsnyder was thinking about looking into it, but otherwise I'll try to dig in later today.

src/view/dataElements/xdmObject.jsx
Aaronius: Here's some other weird behavior. Type "s", then backspace, then "u", then backspace.

filteringproblem2

Aaronius: I think the problem here is that the React-Spectrum ComboBox already has filtering in it (when you type in the input, ComboBox will filter down the options to just those that match), but we're also doing filtering on our side as the user types and then updating the `options` passed into ComboBox. ComboBox is getting all kinds of confused.

I'm not sure, but I think continuing with ComboBox will lead us down a long rabbit hole. I would recommend swapping it out for something like https://git.corp.adobe.com/reactor/extension-adobeanalytics/blob/master/src/view/extensionConfiguration/components/restrictedComboBox.jsx, which leverages Autocomplete but makes it look similar to the ComboBox component. Autocomplete has a `getCompletions` API, which is much more suitable for this type of use case since we won't be competing with the component on who's doing the filtering.

jonsnyder: CompletionComboBox from the analytics extension is closer to what we need here. I am going to try and see if that fixes the issues.

src/view/dataElements/xdmObject.jsx
```diff
@@ -140,63 +177,78 @@ const XdmObject = ({
...
+  const [sandboxes, setSandboxes] = useState();
+  const [schemasMeta, setSchemasMeta] = useState([]);
+  const [schemaOptions, setSchemaOptions] = useState([]);
+  const [schemaSearchInProgress, setSchemaSearchInProgress] = useState();
+  const [schemasMetaSearch, setSchemasMetaSearch] = useState();
   const [schemasMetaStatus, setSchemasMetaStatus] = useState();
-  const [selectedSandboxMeta, setSelectedSandboxMeta] = useState();
+  const [selectedSandbox, setSelectedSandbox] = useState();
   const [selectedSchemaMeta, setSelectedSchemaMeta] = useState();
+  const [showComboMenu, setShowComboMenu] = useState();
```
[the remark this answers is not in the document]
garyr: I think this has been addressed with the move to the `CompletionComboBox` component. For now we have removed the search progress

src/view/dataElements/xdmObject.jsx
garyr: I agree that this should probably be enhanced. I would like to address this with an upcoming change that will likely involve refactoring the sandbox feature into some shared code

src/view/dataElements/xdmObject.jsx
garyr: I think this is addressed with the `CompletionComboBox`

src/view/dataElements/xdmObject.jsx
```diff
@@ -61,19 +77,33 @@ const XdmObject = ({
...
+    debounce(({ sandboxName, search }) => {
+      setSchemaSearchInProgress(true);
+      return fetchSchemasMeta({
+        orgId: initInfo.company.orgId,
+        imsAccess: initInfo.tokens.imsAccess,
+        sandboxName,
+        search
+      }).then(response => {
+        if (!response.results.length) {
+          return;
```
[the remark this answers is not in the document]
garyr: No longer needed

src/view/dataElements/xdmObject.jsx
```diff
@@ -61,19 +77,33 @@ const XdmObject = ({
...
+    []
+  );
 
   return (
     <div>
       <div className="u-gapTop">
         <InfoTipLayout
           tip={
-            sandboxesMeta.disabled && sandboxesMeta.disabled === true
+            sandboxes.disabled && sandboxes.disabled === true
```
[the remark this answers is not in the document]
garyr: Thanks. I made your suggested change

jonsnyder (approved):

Aaronius (approved):
* If I click the dropdown button, then click somewhere outside the combobox, it automatically selects the first item.
* The filtering is case sensitive when it should probably be case insensitive.
* Select a schema. After that, clear out the search field, then click outside of the combobox. I see this shown:

image

The looks like something went wrong, but really the user just deselected the schema they had selected. I would recommend we just down show the alert in that case.

I think case sensitivity would be great to fix before releasing, but we can probably get away with not addressing the other points.

I would recommend squashing when you hit the merge button.

Aaronius:
Thanks for working through this Gary and Jon. I wrote my last comment in a hurry and didn't even preface it by saying "These are a few issues I noticed while poking around." Anyway, it's looking good! Thanks again!

Aaronius:
@garyr, I looked into the TestCafe hanging problem you mentioned. When I was messing around with it, I was able to get it to hang once out of about 20 times of trying different things. I may just not have the reproduction steps down, but anyway, it seemed quite rare. From what I saw, it looked like it's probably more of a bug between TestCafe and the browser than the code we manage. I'm not sure that it's isolated to your branch either.

Anyway, I went ahead and upgraded TestCafe in case that may have any improvement. I didn't see anything that specifically related in the release notes, but there may be something there. I also merged master into this branch because Nina merge a PR that fixed a functional test and I wanted to see all tests passing on this branch.

Anyway, I think you're good to merge whenever you feel ready.

test/functional/helpers/spectrum.js
```diff
@@ -112,15 +112,40 @@ const createExpectDisabled = selector => async () => {
...
     return {
       expectError: createExpectError(selector),
       expectNoError: createExpectNoError(selector),
       expectValue: createExpectValue(selector),
       async expectSelectedOptionLabel(label) {
         await switchToIframe();
+        // There's a bug in TestCafe that incorrectly logs a warning
+        // if this line is formatted differently.
+        // https://github.com/DevExpress/testcafe/issues/5449
+        // prettier-ignore
```
Aaronius, line 144: This was a bug that was introduced in one of the newer versions of TestCafe. It's supposed to be fixed in the next stable release.

test/functional/dataElements/xdmObject.spec.js
```diff
@@ -54,14 +54,7 @@ const expectSettingsToContainData = async data => {
...
   // tests fail if the version we were asserting were hard-coded in the test.
   await t.expect(actualSettings.schema.version).match(/^\d+\.\d+$/);
-  await t
-    .expect(actualSettings.data)
-    .eql(
-      data,
-      `Expected data: ${JSON.stringify(data)} Actual data: ${JSON.stringify(
-        actualSettings.data
-      )}`
-    );
```
Aaronius, line 64: The prior version of TestCafe we were using wasn't very good at showing the diff between two objects when we tried to compare them using `.eql`. It would look like this:

image

which wasn't very useful because it didn't show us which property was different. This is why I had to output both serialized objects like this when the test failed so that the developer could at least compare the objects visually. The latest version added an enhancement that shows a better diff of the object:

image

So now we no longer need to log the stringified objects. 🎉

PR #117, opened 2020-11-04
Fix export of clone utility.

Aaronius:
## Description
We're using ES5 to import modules so we have to use ES5 to export the modules (you don't _have_ to, but the process to import the module is a little more involved). In this case, I did an ES6 `export default`, resulting an error being thrown at runtime. This PR switches it to an ES5 export. I've logged a separate issue to add a lint rule so we can catch this type of thing earlier: https://jira.corp.adobe.com/browse/CORE-54412

## Related Issue
It doesn't resolve this issue, but it is related: https://jira.corp.adobe.com/browse/CORE-54412

## Motivation and Context
The send event action was throwing an error.

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/utils/clone.js
+++ b/src/lib/utils/clone.js
@@ -15,4 +15,4 @@ governing permissions and limitations under the License.
  * @param {*} value
  * @returns {any}
  */
-export default value => JSON.parse(JSON.stringify(value));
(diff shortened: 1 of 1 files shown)


ninaceban (approved):
👍 Thank you!

jonsnyder (approved):

PR #119, opened 2020-11-13
CORE-54412: Adds eslint rule to disable default exports

garyr:
## Description

We would like to prevent using default exports in js,jsx files in src/lib. This will ensure better es5 compatibility with launch

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/.eslintrc.js
+++ b/src/lib/.eslintrc.js
@@ -1,7 +1,10 @@
 module.exports = {
   rules: {
     "no-var": "off",
-    "func-names": "off"
+    "func-names": "off""
+    "func-names": "off",
+    "import/no-default-export": 2,
+    "import/no-named-export": 2
   },
   globals: {
     turbine: "readonly"


src/lib/.eslintrc.js
```diff
@@ -1,9 +1,11 @@
...
-    "func-names": "off"
+    "func-names": "off",
+    "import/no-default-export": 2,
+    "import/no-named-export": 2
   },
   globals: {
     turbine: "readonly"
   }
-};
+};
```
Aaronius, line 11: You might want to configure your IDE/editor to insert a line break at the end of files or we might end up with unnecessary git churn.

Aaronius (approved):

src/lib/.eslintrc.js
garyr: Done

jfkhoury (approved):
Thank you @garyr

PR #120, opened 2020-11-17
Use Alloy's NPM Library instead of copying to runAlloy.js

jonsnyder:
## Description

Use the Alloy NPM library in the launch extension.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-52501

## Motivation and Context

This keeps us from having to manually copy the alloy code to the launch extension.

## Types of changes

## Checklist:


package.json
```diff
@@ -34,6 +34,7 @@
     "test"
   ],
   "dependencies": {
+    "@adobe/alloy": "../alloy",
```
jfkhoury, line 37: Is `../alloy` needed because we are testing locally still?

jfkhoury (approved):
Looks good.

src/lib/instanceManager/injectWrapOnBeforeEventSend.js
```diff
@@ -0,0 +1,37 @@
...
+  content.xdm.implementationDetails.version = `${content.xdm.implementationDetails.version}+${version}`;
+  if (onBeforeEventSend) {
+    const contentClone = clone(content);
+    try {
+      onBeforeEventSend(contentClone);
+      copyTo(content.xdm, contentClone.xdm);
+      copyTo(content.data, contentClone.data);
+    } catch (e) {
+      turbine.logger.error(e);
+    }
```
Aaronius, line 35: I think you're doing this cloning and try-catching because of [how Alloy handles errors coming from the `onBeforeEventSend` callback (it swallows the error and send the unmodified content)](https://github.com/adobe/alloy/blob/8f9b0975f9a5794964e242d6c51c9804e9d0a40b/src/core/createEvent.js#L63-L85)? I'm of the opinion that we should modify the way we handle errors coming from `onBeforeEventSend`, as I've described [here](https://github.com/adobe/alloy/pull/646#discussion_r543520623). I'd like to hear what you think about this.

If anything, a comment here describing why we need this cloning and try-catching would be useful. Also, I notice we're logging an error here if the customer's callback throws an error, but if Alloy were to log an error when the callback throws an error (which it probably should if we stick with the way Alloy is handling callback errors), then I don't think it would be necessary to log an error here, right?

Aaronius (commented):
I'd like to talk through the [questions/comments about the API that I added to the documentation PR](https://git.corp.adobe.com/AdobeDocs/experience-platform.en/pull/1300/files#r3372583). Otherwise, this looks good.

src/lib/instanceManager/injectWrapOnBeforeEventSend.js
jonsnyder: Yes, you are right, and I should add a comment here. I can't remember the original reason why we did the error catching like we do. I agree that we should probably just not send the hit in that case. If a customer wants to they can wrap their own code in a try catch and implement this logic. We should discuss this together though.

The reason why I'm not just letting the error go through is because that would cause Alloy to use the unmodified version of xdm thus removing the extension version.

.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 dist
+temp
```
liljenback, line 3: Shouldn't this be cleaned up by the build instead of being left?

scripts/helpers/build.js
```diff
@@ -46,27 +71,37 @@ module.exports = (options = {}) => {
...
+  // The package resolution in launch does not follow dependencies on npm packages, so we need to build our
+  // own Alloy file here.
+  // I tried having rollup run Babel, but it didn't respect the .browserslistrc located in the src/lib directory.
+  const alloyPromise = fsPromises
+    .mkdir(tempDir)
+    .catch(() => undefined)
+    .then(() => fsPromises.copyFile(browserslistrcFile, browserslistrcTempFile))
+    .then(() => run("rollup", ["-c"]))
+    .then(() => run("babel", [alloyTempFile, "--out-dir", libOutDir]))
+    .then(() => toPromise(callback => rimraf(tempDir, callback)));
```
liljenback, line 104: Have a catch here to make sure to delete tempdir?

scripts/helpers/build.js
```diff
@@ -50,27 +75,37 @@ module.exports = (options = {}) => {
...
+  // The package resolution in launch does not follow dependencies on npm packages, so we need to build our
+  // own Alloy file here.
+  // I tried having rollup run Babel, but it didn't respect the .browserslistrc located in the src/lib directory.
+  const alloyPromise = fsPromises
+    .mkdir(tempDir)
+    .catch(() => undefined)
+    .then(() => fsPromises.copyFile(browserslistrcFile, browserslistrcTempFile))
+    .then(() => run("rollup", ["-c"]))
+    .then(() => run("babel", [alloyTempFile, "--out-dir", libOutDir]))
+    .finally(() => toPromise(callback => rimraf(tempDir, callback)));
```
Aaronius, line 108: My understanding is that Babel doesn't attempt to transpile anything in node_modules by default. That might explain why having rollup run babel without being inside the temp directory wasn't working.

Babel talks about transpiling node_modules here: https://babeljs.io/docs/en/configuration#whats-your-use-case
We might need to change our .babelrc to babel.config.json. We probably should anyway.

Whether we try to do this now or later is up to you.

Aaronius (approved):
🚀

scripts/helpers/build.js
jonsnyder: Are you saying we wouldn't need to use rollup if babel could just transform the files in node_modules?

src/lib/instanceManager/createInstanceManager.js
```diff
@@ -10,51 +10,42 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
   instancesSettings.forEach(
     ({
       name,
       edgeConfigId,
       stagingEdgeConfigId,
       developmentEdgeConfigId,
+      onBeforeEventSend,
       ...options
     }) => {
+      const instance = createInstance({ instanceName: name });
```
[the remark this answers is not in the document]
jonsnyder: I like name better.

scripts/helpers/build.js
Aaronius: No, I think I'm just addressing your comment:

> I tried having rollup run Babel, but it didn't respect the .browserslistrc located in the src/lib directory.

I got the feeling that you had to deal with creating the `temp` directory and running `rollup` and `babel` as separate operations because you couldn't get rollup to run babel while properly transpiling Alloy from node_modules. I think if we change `.babelrc` to `babel.config.json`, then rollup running babel might properly transpile the code, which I think would allow us to also not need this `temp` directory. In other words, my thought is that it's not that babel wasn't respecting `.browserlistrc`, but that babel wasn't running on Alloy from node_modules at all.

I certainly could have misunderstood though.

src/lib/instanceManager/createInstanceManager.js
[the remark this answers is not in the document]
jonsnyder: https://github.com/adobe/alloy/pull/675

Aaronius:
Woohoo! Are we good to merge @jonsnyder?

jonsnyder:
I'm releasing alloy 2.4.0-alpha.0 to NPM right now, and then I'll change the package.json reference here.

PR #123, opened 2021-02-09
Add support for setting Adobe 2.0 consent.

jonsnyder:
## Description

This change allows customers to specify Adobe 2.0 consent objects to the Set Consent action. When the standard is set to Adobe, the UI shows "version", and "value". If "1.0" is typed into the version, the in/out/dataelement options are shown. Otherwise a data element can be specified for the value.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-1074

## Motivation and Context

The new Adobe 2.0 standard uses an XDM object to specify consent and preferences. The previous Adobe 1.0 standard expected an object like { general: "in" }. At this time we didn't want to create a custom UI for the consent XDM, and so it can only be specified with a data element.

## Types of changes

## Checklist:

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -223,6 +223,14 @@
                     {
                       "type": "object",
                       "properties": {
+                        "standard": { "type": "string", "enum": ["Adobe"] },
+                        "version": { "type": "string" },
+                        "value": { "type": "string", "pattern": "^%[^%]+%$" }
+                      }
+                    },
+                    {
+                      "type": "object",
+                      "properties": {
                         "standard": { "type": "string", "enum": ["IAB TCF"] },
                         "version": { "type": "string" },
                         "value": { "type": "string" },
--- a/src/lib/.eslintrc.js
+++ b/src/lib/.eslintrc.js
@@ -2,7 +2,6 @@ module.exports = {
   rules: {
     "no-var": "off",
     "func-names": "off",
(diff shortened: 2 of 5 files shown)


Aaronius (approved):
This looks good. I was going to run it in the sandbox locally to do some manual validation, but I keep running into errors when installing dependencies and running the sandbox. I got past some of the errors , but there are more I need to get through. I think it has to do with my version of node and npm (I'm on node 14.15.3, npm 6.14.9) and I think we'll need to upgrade our dependencies to fix it. Anyway, you don't need to do anything about it--just making you aware.

extension.json
```diff
@@ -220,6 +220,14 @@
                       },
                       "additionalProperties": false
                     },
+                    {
+                      "type": "object",
+                      "properties": {
+                        "standard": { "type": "string", "enum": ["Adobe"] },
+                        "version": { "type": "string" },
```
liljenback, line 227: This won't have a pattern you can verify against?

ninaceban (approved):
👍

jfkhoury (approved):
✌️

extension.json
jonsnyder: This is a good idea, but unfortunately, it is too late to add validation here. We can't make old settings invalid now.

PR #126, opened 2021-02-23
PDCL-1239 - Return a better error in XDM Object if token invalid

garyr:
## Description

For users of the launch extension who do not have AEP access, there is a confusing error message that is displayed about not being able to load sandboxes. This change replaces the error message to something more human readable:

```Your user account is not enabled for AEP access. Please contact your organization administrator.```

## Types of changes

## Checklist:

### CODE REVIEW

**Review by jfkhoury — APPROVED** [review]
Awesome @garyr ✌️

File: src/view/dataElements/xdmObject/helpers/parseResponseJson.js [hunk 1]
```diff
@@ -0,0 +1,21 @@
...
+    response.json().then(body =>
+      resolve({
+        status: response.status,
+        ok: response.ok,
+        json: () => {
+          return body;
+        }
+      })
+    );
+  });
```
Aaronius, line 20: No need to create a new promise here. Try this.
  
```
return response.json().then(body => {
  return {
    status: response.status,
    ok: response.ok,
    json() {
      return body;
    }
  }
})
```

I haven't looked at the rest of this PR yet, but you'll need to make sure we're appropriately handling a `response.ok` of `false`.

test/functional/dataElements/xdmObject.spec.js
```diff
@@ -103,6 +105,26 @@ test("initializes form fields with individual object attribute values", async ()
   await stringEdit.expectValue("Adobe");
 });
 
+test.only("ensures invalid token error", async () => {
+  // temporarily remove sandboxes mock
+  await t.removeRequestHooks(platformMocks.sandboxes);
+  // replace with unauthorized mock
+  await t.addRequestHooks(platformMocks.unauthorized);
+  await initializeExtensionView();
+  await genericAlert.expectTitle("Your access token appears to be invalid.");
```
Aaronius, line 114: Hmm...I don't think this test is actually working. What made me suspicious is that the error message should show up in the body of the alert and not the header. When I try to run the test on my machine, the test passes before the UI even loads, which shouldn't happen. I can also change the expected message and it still passes as well. Let me investigate this.

Aaronius (commented):
Thanks for this improvement Gary! Something's up with your tests. You still have quite a bit of work to do on https://github.com/adobe/alloy/pull/660, so let me take this PR from here to get us over the finish line.

test/functional/helpers/spectrum.js
```diff
@@ -237,7 +240,17 @@ const componentWrappers = {
     return {
       async expectTitle(title) {
         await switchToIframe();
-        await selector.find(".spectrum-Alert-header").withText(title);
```
Aaronius, line 240: This is merely a selector, so we weren't ever doing an assertion that an element matching the selector actually exists, which was allowing the test to pass when it shouldn't have.

garyr: Thanks

garyr (commented):
Looks really good. Thank you for working on this

Aaronius (approved):

PR #128, opened 2021-02-24
PDCL-1122 Fixes a cross origin error when running inside Launch UI.

Aaronius:
## Description
Fixes a cross origin error when running inside Launch UI. This cross origin error would only show up when the extension view was running inside the Launch UI because the Launch UI is in a different origin that the extension view. The error, however, never actually impacted the extension view. The code causing the cross-origin error existed to support the React dev tools browser extension. I used this extension for a moment while I was building the XDM Object data element, but I don't think anyone is using it anymore. I removed the offending code, but if someone needs to use the React dev tools extension, we could handle this in other ways.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-1122

## Motivation and Context
Users were reporting this error, but it's always a red herring to the real underlying problem they're facing.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/identityMap.html
+++ b/src/view/dataElements/identityMap.html
@@ -1,12 +1,12 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <title>Extension View</title>
-</head>
-<body>
-<div id="root"></div>
-<script src="https://assets.adobedtm.com/activation/reactor/extensionbridge/extensionbridge.min.js"></script>
-<script src="identityMap.jsx"></script>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <title>Extension View</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script src="https://assets.adobedtm.com/activation/reactor/extensionbridge/extensionbridge.min.js"></script>
+    <script src="./identityMap.jsx"></script>
+  </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/view/dataElements/identityMap.html
```diff
@@ -1,12 +1,12 @@
...
+  <head>
+    <meta charset="UTF-8">
+    <title>Extension View</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script src="https://assets.adobedtm.com/activation/reactor/extensionbridge/extensionbridge.min.js"></script>
+    <script src="./identityMap.jsx"></script>
+  </body>
 </html>
```
Aaronius, line 12: These two HTML files were fine the way they were, but I decided to make them consistent with all our other HTML files.

src/view/dataElements/xdmObject.html
```diff
@@ -3,15 +3,10 @@
   <head>
     <meta charset="UTF-8">
     <title>Extension View</title>
-    <script>
-      if (window.parent !== window) {
-        window.__REACT_DEVTOOLS_GLOBAL_HOOK__ = window.parent.__REACT_DEVTOOLS_GLOBAL_HOOK__;
-      }
```
Aaronius, line 9: This was the offending code.

jfkhoury (approved):
🥇

jonsnyder (approved):

PR #131, opened 2021-03-25
PDCL-4413 Fix validation for required fields to take into considerati…

Aaronius:
…on whether the parent object has been populated.

## Description
Consider a schema like this:

image

In this schema, the `_acopprod3` field is NOT marked as required. Neither is `_acopprod3.reviewText`. However, `_accopprod3.productSku`, `_accopprod3.rating`, and `_accopprod3.reviewerId` ARE marked as required.

If you go to create a data element using the XDM Object data element type, select this schema, don't any values to any fields, then try to save, the XDM Object view prevents us from saving the data element and shows validation errors for `_accopprod3.productSku`, `_accopprod3.rating`, and `_accopprod3.reviewerId` fields.

This behavior is incorrect, because if I don't use the XDM Object data element and send data without those fields, the data passes validation within AEP. AEP's behavior makes sense, because they follow the JSON-schema standard, and the way the JSON-schema standard works is it only flags a required field as invalid if the field's parent object is provided in the JSON being validated. 

This PR changes the validation logic to match how required fields are being validated in Adobe Experience Platform.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-4413

## Motivation and Context
The extension's validation should match Adobe Experience Platform validation for required fields.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/xdmObject/helpers/array/validate.js
+++ b/src/view/dataElements/xdmObject/helpers/array/validate.js
@@ -34,11 +34,21 @@ export default ({
 
   if (items) {
     const itemErrors = items.map(itemFormStateNode => {
-      return validate({
+      let itemIsPopulated = false;
+      let itemError = validate({
         formStateNode: itemFormStateNode,
-        isParentAnArray: true,
-        notifyParentOfDataPopulation: confirmDataPopulatedAtCurrentOrDescendantNode
+        confirmDataPopulatedAtCurrentOrDescendantNode() {
+          itemIsPopulated = true;
+          confirmDataPopulatedAtCurrentOrDescendantNode();
+        }
       });
+      if (!itemIsPopulated && !itemError) {
+        itemError = {
+          value:
+            "Items within arrays must not be empty. Please populate or remove the item."
+        };
+      }
+      return itemError;
     });
     if (itemErrors.filter(error => error).length) {
       return { items: itemErrors };
--- a/src/view/dataElements/xdmObject/helpers/object/validate.js
+++ b/src/view/dataElements/xdmObject/helpers/object/validate.js
@@ -13,6 +13,7 @@ governing permissions and limitations under the License.
 import { WHOLE } from "../../constants/populationStrategy";
 import isFormStateValuePopulated from "../isFormStateValuePopulated";
 import singleDataElementRegex from "../../../../constants/singleDataElementRegex";
+import { NONE } from "../../constants/autoPopulationSource";
 
 export default ({
   formStateNode,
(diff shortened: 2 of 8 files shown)


src/view/dataElements/xdmObject/helpers/array/validate.js
```diff
@@ -34,11 +34,21 @@ export default ({
...
+          confirmDataPopulatedAtCurrentOrDescendantNode();
+        }
       });
+      if (!itemIsPopulated && !itemError) {
+        itemError = {
+          value:
+            "Items within arrays must not be empty. Please populate or remove the item."
+        };
+      }
+      return itemError;
```
Aaronius, line 51: This validation of empty items within arrays was moved here from xdmObject/helpers/validate.js. It really belonged here because it's an array-specific validation.

src/view/dataElements/xdmObject/helpers/object/validate.js
```diff
@@ -33,20 +34,62 @@ export default ({
...
+              propertyErrorsMemo[propertyName] = {
+                value: "This is a required field and must be populated."
+              };
+            }
+          }
+          return propertyErrorsMemo;
+        },
+        propertyErrors
+      );
+    }
```
Aaronius, line 92: This is the code that fixes the issue.

src/view/dataElements/xdmObject/helpers/validate.js
```diff
@@ -10,63 +10,37 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-  if (
-    isParentAnArray &&
-    !isAutoPopulated &&
-    !isPopulatedAtCurrentOrDescendantNode
-  ) {
-    return {
-      value:
-        "Items within arrays must not be empty. Please populate or remove the item."
-    };
-  }
```
Aaronius, line 59: This logic got moved into the array-specific `validate` module. By moving it, we were able to get rid of `isParentAnArray`.

src/view/dataElements/xdmObject/helpers/validate.js
```diff
@@ -10,63 +10,37 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-
-  if (
-    schema.isRequired &&
-    !isAutoPopulated &&
-    !isPopulatedAtCurrentOrDescendantNode
-  ) {
-    return { value: "This is a required field and must be populated." };
-  }
-
-  return undefined;
```
Aaronius, line 69: This logic got moved to the object-specific `validate` module and enhanced to fix the bug we were seeing. This should have been in the object-specific `validate` module from the beginning, but I wasn't smart enough when I initially wrote it to realize that that was the case. The concept of required fields as far as XDM schemas go is specific to object properties, so it makes sense that this would be in the object-specific `validate` module.

test/functional/dataElements/xdmObject.spec.js
```diff
@@ -315,24 +315,6 @@ test("initializes form fields with whole array value", async () => {
...
-  await arrayEdit.addItem();
-  await arrayEdit.clickItem(0);
-  await arrayEdit.enterValue("%item1%");
-  await xdmTree.toggleExpansion("industries");
-
-  await extensionViewController.expectIsNotValid();
-  await xdmTree.expectIsValid("Item 1");
-  await xdmTree.expectIsNotValid("Item 2");
-});
-
```
Aaronius, line 335: I created a separate spec encompassing validation and moved this test there.

test/functional/dataElements/xdmObjectValidation.spec.js
```diff
@@ -0,0 +1,81 @@
...
+  await initializeExtensionView();
+  await selectSchemaFromSchemasMeta();
+  await xdmTree.toggleExpansion("_unifiedjsqeonly");
+  await xdmTree.toggleExpansion("customer");
+  // mailingAddress and emailAddress are siblings.
+  await xdmTree.toggleExpansion("mailingAddress");
+  await xdmTree.click("city");
+  await stringEdit.enterValue("San Jose");
+  await extensionViewController.expectIsNotValid();
+  await xdmTree.expectIsNotValid("emailAddress");
```
Aaronius, line 80: I added this `customer` node to the XDM schema. It contains this `emailAddress` field which I marked as required.

src/view/dataElements/xdmObject/helpers/object/validate.js
Aaronius: As a side note, I'm much of a fan of `reduce` for the two cases above as I feel like it makes the code unnecessarily complex and difficult to follow. I did it anyway because that seems to follow our [convention](https://github.com/adobe/alloy/wiki/Conventions#reduce-instead-of-foreach).

src/view/dataElements/xdmObject/helpers/validate.js
```diff
@@ -10,63 +10,37 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  const { schema } = formStateNode;
 
   const errors = getTypeSpecificHelpers(schema.type).validate({
     formStateNode,
-    confirmDataPopulatedAtCurrentOrDescendantNode,
+    // By using "once", we ensure that the parent is notified that
+    // data is populated at the current or descendant node at most
+    // a single time. This is primarily for optimization but it's
+    // also easier to reason about from the parent node's perspective.
+    confirmDataPopulatedAtCurrentOrDescendantNode: once(
```
Aaronius, line 37: We were doing this `once` logic manually [here](https://github.com/adobe/reactor-extension-alloy/pull/131/files#diff-a20847d2c2ec831d3f5b1825e7ffca85e7d6ee51d9d20137a7302963f25b92f2L33-L38). I decided to just leverage the npm package.

jonsnyder (approved):
Thanks for cleaning up parts of the code as you touch it.

test/functional/dataElements/xdmObjectValidation.spec.js
```diff
@@ -0,0 +1,81 @@
+/*
```
jfkhoury, line 1: Nice Test!

jfkhoury (approved):

PR #134, opened 2021-05-05
Add data field for sendEvent

jonsnyder:
## Description

Add a field on sendEvent to specify a data element for data.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-5942

## Motivation and Context

This is to support sending target profile attributes inside data.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] All tests pass and I've made any necessary test changes.
- [x] I've updated the schema in extension.json or no changes are necessary.

PR #137, opened 2021-05-10
PDCL-6009 Upgrade Send Event view to React-Spectrum 3 and React Hook Form.

Aaronius:
## Description
This upgrades the Send Event action view from React-Spectrum 2 to React-Spectrum 3 as well from Formik to React Hook Form. It also lays the foundation for upgrading the remaining views. This PR is opened against a new `react-spectrum` branch where we'll stage all React-Spectrum upgrade changes until we get other views upgraded and then we'll merge the `react-spectrum` to `main` and release.

The other views that are not the Send Event view continue to function properly with React-Spectrum 2 and Formik.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-6009

## Motivation and Context

For React-Spectrum:
* It has some valuable ComboBox behaviors that we could use to support more than 100 schemas and edge configurations.
* It has some other components that will come in handy for supporting XDM merging.
* It will allow building the extension outside of Adobe's network.
* It's also an inevitable progression of our UI.

For React Hook Form:
* We've learned from the Launch team that Formik can cause performance issues with React-Spectrum 3. Also, React Hook Form seems like a step up from Formik and seems to be simplifying some code for us.

## Types of changes

## Checklist:


src/view/actions/sendEvent.jsx
```diff
@@ -60,22 +55,22 @@ const getDecisionScopesFromFormState = values => {
 const getInitialDecisionScopesData = decisionScopes => {
   if (Array.isArray(decisionScopes)) {
     return {
-      decisionsInputMethod: decisionScopesOptions.CONSTANT,
-      decisionScopesArray: decisionScopes,
+      decisionsInputMethod: CONSTANT,
+      decisionScopesArray: decisionScopes.map(scope => ({ value: scope })),
```
Aaronius, line 59: One downside of React Hook Form is that when dealing with editing arrays of items, the items must be objects (and not just strings, like we had). This is a current limitation of React Hook Formik because the object is used to store a key that [helps with performance when re-rendering lists of items](https://reactjs.org/docs/lists-and-keys.html#keys).

src/view/actions/sendEvent.jsx
```diff
@@ -139,20 +134,20 @@ const getSettings = ({ values }) => {
 };
 
 const validationSchema = object().shape({
-  xdm: string().matches(
-    singleDataElementRegex,
-    "Please specify a data element"
-  ),
+  xdm: string().matches(singleDataElementRegex, {
+    excludeEmptyString: true,
```
Aaronius, line 138: We initialize the form with an empty string for the `xdm` field. Upon validation, Formik would massage the data before passing it to validation such that the value being validated were not an empty string, which allowed this validation to pass. React Hook Form, on the other hand, doesn't attempt to massage the data and therefore passes an empty string to yup, which wouldn't normally pass validation because it doesn't match the regex. `excludeEmptyString` allows the validation to pass when the value is an empty string, which is what we want in this case.

src/view/actions/sendEvent.jsx
```diff
@@ -10,48 +10,43 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 import singleDataElementRegex from "../constants/singleDataElementRegex";
-import "./sendEvent.styl";
-import InfoTipLayout from "../components/infoTipLayout";
+// import "./sendEvent.styl";
 import DecisionScopesComponent from "../components/decisionScopesComponent";
-
-const decisionScopesOptions = {
-  CONSTANT: "constant",
-  DATA_ELEMENT: "dataElement"
-};
```
Aaronius, line 34: I broke these out into an enum file so that DecisionScopesComponent could access them directly instead of SendEvent sending them down through props.

src/view/actions/sendEvent.jsx
```diff
@@ -10,48 +10,43 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+import { CONSTANT, DATA_ELEMENT } from "../constants/decisionScopesInputMethod";
 
 const filterDecisionScopes = scopes => {
-  return scopes.filter(s => s !== "");
+  return scopes.filter(scope => scope.value !== "");
```
Aaronius, line 32: One downside of React Hook Form is that when dealing with editing arrays of items, the items must be objects (and not just strings, like we had). This is a current limitation of React Hook Formik because the object is used to store a key that [helps with performance when re-rendering lists of items](https://reactjs.org/docs/lists-and-keys.html#keys).

src/view/actions/resetEventMergeId.jsx
```diff
@@ -16,9 +16,9 @@ import { object, string } from "yup";
 import Textfield from "@react/react-spectrum/Textfield";
 import FieldLabel from "@react/react-spectrum/FieldLabel";
 import "@react/react-spectrum/Form"; // needed for spectrum form styles
-import render from "../render";
+import render from "../spectrum2Render";
```
Aaronius, line 19: Some of these files, like `render.jsx`, are now different between the React-Spectrum-2/Formik views and the React-Spectrum-3/React-Hook-Form views. To keep a clear distinction, I named some of the files accordingly. Once we're fully off React-Spectrum-2/Formik, we'll probably rename them back to something more generic like `render.jsx`.

src/view/components/decisionScopesComponent.jsx
```diff
@@ -10,116 +10,99 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+      {decisionsInputMethod === CONSTANT && (
+        <div className="FieldSubset">
+          {decisionScopeFields.map((field, index) => {
+            return (
+              <div className="u-gapBottom" key={field.id}>
+                <TextField
+                  data-test-id={`scope${index}Field`}
+                  name={`decisionScopesArray.${index}.value`}
+                  width="size-5000"
+                  defaultValue={field.value}
```
Aaronius, line 71: This is kind of another gotcha with React Hook Form. When dealing with field arrays, you have to set `defaultValue` because of how it optimizes rendering of items. https://github.com/react-hook-form/react-hook-form/discussions/5048

src/view/global.styl
```diff
@@ -14,12 +14,17 @@ governing permissions and limitations under the License.
...
+html
 body,
 #root,
 .react-spectrum-provider {
-  box-sizing: border-box;
-  height: 100%;
+  display: flex;
+  flex-direction: column;
+  flex: 1
 }
```
Aaronius, line 28: Previously the body would never grow larger than the height of the iframe. This resulted in no padding being displayed at the bottom of some views when scrolled to the bottom:

image


My changes fix that.

jonsnyder (approved):
I'm excited for these changes. I love that the components now include the label. This will make the markup less complicated.

jonsnyder:
I just merged https://github.com/adobe/reactor-extension-alloy/pull/134, which adds a data field. We will need to merge main at some point.

src/view/components/dataElementSelector.jsx
```diff
@@ -0,0 +1,71 @@
...
+*/
+
+import React from "react";
+import PropTypes from "prop-types";
+import { useFormContext } from "react-hook-form";
+import Data from "@spectrum-icons/workflow/Data";
+import { Button } from "@adobe/react-spectrum";
+import "./dataElementSelector.styl";
+
+const DataElementSelector = ({ children, name, augmentValue }) => {
```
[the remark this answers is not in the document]
Aaronius: Yes! I can, thank. you.

src/view/components/dataElementSelector.jsx
```diff
@@ -0,0 +1,71 @@
...
+  };
+  return (
+    <div className="u-flex">
+      <div>{children}</div>
+      <Button
+        variant="secondary"
+        isQuiet
+        onPress={openDataElementSelector}
+        aria-label="Select data element"
+        UNSAFE_className={
```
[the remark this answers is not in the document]
Aaronius: It's the only way you can add a class to React-Spectrum components. More info here: https://react-spectrum.adobe.com/react-spectrum/styling.html#escape-hatches

src/view/components/dataElementSelector.jsx
```diff
@@ -0,0 +1,71 @@
...
+          adjustForLabel ? "DataElementSelector-buttonAdjustedForLabel" : ""
+        }
+        minWidth={0}
+      >
+        <Data />
+      </Button>
```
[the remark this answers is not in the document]
Aaronius: Not while staying Spectrum-compliant.

test/functional/helpers/spectrum3.js
```diff
@@ -0,0 +1,264 @@
...
+      async enterSearch(text) {
+        await switchToIframe();
+        await t.typeText(selector, text).pressKey("enter");
+      },
+      async clear() {
+        await switchToIframe();
+        await t.selectText(selector).pressKey("delete");
+      }
+      // expectDisabled: createExpectDisabled(selector.parent().find("button")),
+      // expectEnabled: createExpectEnabled(selector.parent().find("button")),
```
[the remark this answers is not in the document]
Aaronius: I was wanting to wait until I started working on a view spec that actually used them to make sure they work properly. I think I'll delete them for now and add functional ones back once I get to a view spec that uses them.

src/view/actions/sendEvent.jsx
```diff
@@ -174,138 +169,85 @@ const knownEventTypes = [
...
+          <Form>
+            <Picker
+              data-test-id="instanceNameField"
+              name="instanceName"
+              label="Instance"
+              items={getInstanceOptions(initInfo)}
+              width="size-5000"
+            >
+              {item => <Item key={item.value}>{item.label}</Item>}
+            </Picker>
```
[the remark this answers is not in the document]
Aaronius: Yeah, good question. I left it off because there's technically no way for the user to deselect an instance name. On second thought, I think it's probably good to add it, especially since it might help the user to understand that the other fields are optional.

src/view/components/decisionScopesComponent.jsx
```diff
@@ -10,116 +10,99 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
     </div>
   );
 }
 
-DecisionScopesComponent.propTypes = {
-  values: PropTypes.object,
-  options: PropTypes.object
-};
-
 export default DecisionScopesComponent;
```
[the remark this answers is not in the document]
Aaronius: Thanks for finding this. This definitely looks like the React Hook Form issue we discussed.

src/view/actions/sendEvent.jsx
```diff
@@ -60,22 +55,22 @@ const getDecisionScopesFromFormState = values => {
...
     };
   }
   return {
-    decisionsInputMethod: decisionScopesOptions.CONSTANT,
+    decisionsInputMethod: CONSTANT,
     decisionScopesDataElement: "",
-    decisionScopesArray: [""]
+    decisionScopesArray: [{ value: "" }]
   };
 };
```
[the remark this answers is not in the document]
Aaronius: I agree. I've added a commit that moves the decision scopes logic into the decision scopes component file.

Aaronius:
@jonsnyder, as we discussed, I've changed this back to Formik from React Hook Form. I upgraded Formik which allowed us to use some of the new hooks that Formik provides which cleans up some things. I feel good about this.

Aaronius:
Oh, I also resolved the other issues you brought up.

The validation is a little different than what we used to have. Previously, we would run validation on every change event but only show the error on the field if the field had previously been "touched" (or in other words, received focus then lost focus). For example, if I were to start typing in the XDM field and it didn't start with a `%`, you wouldn't see an error on the field until the field lost focus. But, if you went back to the field and started backspacing to clear out all the text and start typing `%something%`, the validation error would go away once the value passed validation regardless of whether the field had lost focus yet.

Anyway, we could do something similar with our new setup, but I kind of want to roll with how we have it for a while and see how it feels. The new setup runs validation on blur but never on change. I happy to hear feedback on this.

src/view/components/hookFormReactSpectrum/picker.jsx
```diff
@@ -29,7 +29,9 @@ const Picker = ({ name, description, width, ...otherProps }) => {
         {...otherProps}
         selectedKey={value}
         onSelectionChange={setValue}
-        onBlur={onBlur}
```
Aaronius: The `onBlur` from Formik expects a native blur event to be passed. Since React-Spectrum doesn't provide the native event, the field wasn't being marked as touched. `setTouched` is designed more for custom components like we have here.

src/view/components/fieldDescriptionAndError.jsx
```diff
@@ -34,8 +34,6 @@ const FieldDescriptionAndError = ({
   } else if (description) {
     messageClassName = "FieldDescriptionAndError-description";
     message = description;
-  } else {
-    return children;
```
Aaronius: You may wonder why I changed this. Until I configured Formik to _not_ validate on change events, validation was happening on every change (keystroke). What I noticed is that if I typed a character that caused the field to go from valid to invalid (or vice-versa), the field would lose focus. That's because, in order to show the error message, we were taking the input out of its position in the DOM and then adding the div wrapper (see below). The fix is to not move the input around in the DOM while the user is typing, which is what this change fixes.

Now, since I made this change, I've turned off validation on change events, so moving the input in the DOM when validity changes is not a problem we'll run into, but I left this change in place in case we change our mind on when we want to validate.

src/view/components/dataElementSelector.jsx
```diff
@@ -17,16 +17,16 @@ import Data from "@spectrum-icons/workflow/Data";
...
-  let adjustForLabel = false;
-  React.Children.forEach(children, child => {
-    if (Object.keys(child.props).includes("label")) {
-      adjustForLabel = true;
-    }
-  });
+  const inputChild = React.Children.toArray(children).find(
+    child => child.props.name
+  );
+  const name = inputChild.props.name;
```
jonsnyder, line 27: Should we add a check for a null inputChild? Probably throw an error more friendly than "cannot get property 'props' of null"

src/view/actions/sendEvent.jsx
```diff
@@ -18,85 +18,40 @@ import {
   Picker,
   TextField,
   Checkbox
-} from "../components/hookFormReactSpectrum";
+} from "../components/formikReactSpectrum";
```
jonsnyder, line 21: Should this be formikReactSpectrum3

src/view/actions/sendEvent.jsx
```diff
@@ -125,26 +83,21 @@ const getSettings = ({ values }) => {
...
+  Object.assign(settings, getDecisionScopesSettings({ values }));
 
   return settings;
 };
 
 const validationSchema = object().shape({
   xdm: string().matches(singleDataElementRegex, {
-    excludeEmptyString: true,
     message: "Please specify a data element"
   }),
```
jonsnyder, line 95: Now that you removed the excludeEmptyString key, you can just pass the message as the second parameter.

src/view/components/dataElementSelector.jsx
Aaronius: Yeah, that's probably a good idea.

src/view/actions/sendEvent.jsx
Aaronius: I didn't put `3` in there because there was no naming collision, but it might be better to do so for clarity until we have react-spectrum 2 cleared out.

PR #139, opened 2021-05-14
PDCL-5957 Fixes issue where error was thrown if sandboxes were returned but without a default.

Aaronius:
## Description
When a customer has access to at least one sandbox but not sandbox configured as the default for the org, the following error would be thrown: `Cannot read property 'name' of undefined`. This fixes that issue.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-5957

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/xdmObject.jsx
+++ b/src/view/dataElements/xdmObject.jsx
@@ -248,14 +248,10 @@ const XdmExtensionView = () => {
           imsAccess: initInfo.tokens.imsAccess
         })
           .then(response => {
-            let sandbox = { name: null };
+            let sandbox;
             // if sandboxes are returned from the api call
             if (response.sandboxes && response.sandboxes.length) {
               setSandboxes(response);
-              // default to the first item in the list
-              sandbox = response.sandboxes.find(
-                _sandbox => _sandbox.isDefault === true
-              );
 
               // if a sandbox exists in settings and is in the list, choose that item
               if (initInfo.settings && initInfo.settings.sandbox) {
@@ -263,8 +259,24 @@ const XdmExtensionView = () => {
                   _sandbox => _sandbox.name === initInfo.settings.sandbox.name
                 );
               }
-              setSelectedSandbox(sandbox);
+
+              // default to the first item in the list
+              if (!sandbox) {
+                sandbox = response.sandboxes.find(
+                  _sandbox => _sandbox.isDefault
+                );
+              }
+
+              if (!sandbox) {
+                [sandbox] = response.sandboxes;
+              }
             }
+
+            if (!sandbox) {
+              sandbox = { name: null };
+            }
+
+            setSelectedSandbox(sandbox);
 
             return sandbox;
           })
--- a/src/view/dataElements/xdmObject.jsx
+++ b/src/view/dataElements/xdmObject.jsx
@@ -272,10 +272,6 @@ const XdmExtensionView = () => {
               }
             }
 
-            if (!sandbox) {
-              sandbox = { name: null };
-            }
-
             setSelectedSandbox(sandbox);
 
             return sandbox;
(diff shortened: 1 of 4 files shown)


src/view/dataElements/xdmObject.jsx
```diff
@@ -248,24 +248,36 @@ const XdmExtensionView = () => {
...
           .then(response => {
-            let sandbox = { name: null };
+            let sandbox;
             // if sandboxes are returned from the api call
             if (response.sandboxes && response.sandboxes.length) {
               setSandboxes(response);
-              // default to the first item in the list
-              sandbox = response.sandboxes.find(
-                _sandbox => _sandbox.isDefault === true
-              );
```
Aaronius, line 258: This is where the issue was. We wouldn't find a sandbox with `isDefault` set to `true`, leaving `sandbox` undefined.

src/view/dataElements/xdmObject.jsx
```diff
@@ -248,24 +248,36 @@ const XdmExtensionView = () => {
...
+                );
+              }
+
+              if (!sandbox) {
+                [sandbox] = response.sandboxes;
+              }
             }
 
+            if (!sandbox) {
+              sandbox = { name: null };
```
Aaronius, line 276: I moved this from a different spot. I don't like this idea of creating a half-baked sandbox object and I'm not sure it's actually necessary. Some observations: 

1. `fetchSandboxes` ensures that if the response body has a `sandboxes` property that's an empty array, that it automatically adds a default sandbox object: https://github.com/adobe/reactor-extension-alloy/blob/c58174792eb61226183fef8f6c3a8d2c1c56202a/src/view/dataElements/xdmObject/helpers/fetchSandboxes.js#L47-L49

2. The code above this line wrapped in `if (response.sandboxes && response.sandboxes.length) {` ensures that if there's at least one sandbox in `response.sandboxes`, that the `sandbox` variable _will_ get set.

Based on those observations, it seems that we would only be creating this half-baked sandbox object if we get a response and the response didn't have a `sandboxes` property on it. I'm not sure when that would occur and I don't see anything in the API docs suggesting that might occur. I also don't see any functional test for it. It might just be dead code. @garyr maybe you can provide some insight here?

Aaronius: I plan on removing the half-baked sandbox object in the react-spectrum upgrade btw.

jonsnyder (approved):

jonsnyder:
Great job cleaning up the tests too!

jfkhoury (approved):
Thank you!

src/view/dataElements/xdmObject.jsx
Aaronius: Sounds like a reasonable explanation. I've added a commit to remove this partial sandbox object. If we do find there's a case for it, we'll solve the issue and add a functional test for it.

test/functional/dataElements/xdmObject.spec.js
```diff
@@ -78,13 +79,8 @@ const initializeExtensionView = async additionalInitInfo => {
...
-  .beforeEach(async () => {
-    // adds a mocked sandboxes response
-    await t.addRequestHooks(platformMocks.sandboxes);
-  })
   .disablePageReloads.page("http://localhost:3000/viewSandbox.html")
-  .meta("requiresAdobeIOIntegration", true)
-  .requestHooks(platformMocks.sandboxes);
+  .meta("requiresAdobeIOIntegration", true);
 
 test("initializes form fields with individual object attribute values", async () => {
```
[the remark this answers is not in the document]
Aaronius: Actually, I don't think we need `platformMocks.sandboxes` at all. All the tests we have work fine with the real sandboxes that come back from Platform and I'd rather be testing with the real sandboxes when possible anyway. I've added a commit that removes `platformMocks.sandboxes`.

Aaronius:
I added a couple commits based on @jonsnyder's feedback. Feel free to review. Otherwise I'll merge soonish.

PR #141, opened 2021-05-25
Adding Subforms

jonsnyder:
## Description

This PR separates the Formik setup and the ExtensionView setup. This enabled you to create multiple Formik forms on a page. This will help us to create re-useable form components that can all do their own getInitialValues, getSettings, and validation schema independent from each other. I applied these changes to the sendEvent view.

## Types of changes

## Checklist:


src/view/components/decisionScopes.jsx
```diff
@@ -152,12 +69,114 @@ export const getSettings = ({ values }) => {
...
+      getSettings={getSettings}
+      validationSchema={validationSchema}
+      render={({ formikProps }) => {
+        const {
+          decisionsInputMethod,
+          decisionScopesArray
+        } = formikProps.values;
+
+        return (
+          <Fragment>
```
Aaronius, line 98: You may know this already, but I think you can just do `<>` instead of importing `Fragment`. Either one works for me though.

src/view/components/formikReactSpectrum3/textField.jsx
```diff
@@ -28,7 +28,9 @@ const TextField = ({ name, description, width, ...otherProps }) => {
       <ReactTextField
         {...otherProps}
         value={value}
-        onChange={setValue}
+        onChange={newValue => {
+          setValue(newValue);
+        }}
```
Aaronius, line 33: Looks like you can revert this unless you want it for debugability or something.

Aaronius (approved):
Love it. Looking forward to using it!

Aaronius:
From looking at the code, this excludes unmounted subforms from the validate/getSettings process, right? There's no way to include unmounted subforms currently? I think that's fine until we see we need it.

src/view/components/extensionViewForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+import PropTypes from "prop-types";
+import wrapValidateWithErrorLogging from "../utils/wrapValidateWithErrorLogging";
+import ExtensionViewContext from "./extensionViewContext";
+
+const ExtensionViewForm = ({
+  initialValues,
+  getSettings,
+  validate,
+  validationSchema,
+  render
```
Aaronius, line 12: We talked about adding another prop here for validating things that aren't Formik-related. Did you want to do that? I can do it when I start to use this on my branch if you want.

jonsnyder: You are welcome to add that.

jonsnyder:
I worked on supporting un-mounted subforms. I got it working, but I was unhappy with the result so I removed it from this PR. I have a better solution in mind that I will submit as a separate PR. After working on it a bit, I realized there are a few concerns with transient forms:

1. Getting the settings when the form isn't visible.
2. Preserving the error state of a form when it is not visible.
3. Reseting the form state when the subform becomes visible again.

In some cases like an accordion, all three concerns apply. In other cases like choosing between using a data element and a form, only the 3rd concern applies. For a modal dialog, I don't think any of these apply. Instead, you'd want something that could get the settings before the modal closed, and you would only allow the modal to be closed if the form was valid. After the modal closed, you could put the resulting settings into the formikState on the main form.

PR #142, opened 2021-05-29
Fix test failures due to Formik upgrade.

Aaronius:
## Description
In https://github.com/adobe/reactor-extension-alloy/commit/7d4a9408c976e3bf9c353a7525bf749db697a8a0, I upgraded Formik, which broke some functional tests. Something changed in Formik between the version we were using and the new version. For field values that are empty strings, Formik used to report an empty string. Now it converts empty strings to undefined before passing them to validation. Related: https://github.com/formium/formik/issues/805

This PR fixes related test failures.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/actions/resetEventMergeId.jsx
+++ b/src/view/actions/resetEventMergeId.jsx
@@ -22,6 +22,7 @@ import ExtensionView from "../components/spectrum2ExtensionView";
 import singleDataElementRegex from "../constants/singleDataElementRegex";
 import InfoTipLayout from "../components/infoTipLayout";
 import "./resetEventMergeId.styl";
+import { DATA_ELEMENT_REQUIRED } from "../constants/validationErrorMessages";
 
 const getInitialValues = ({ initInfo }) => {
   const { eventMergeId = "" } = initInfo.settings || {};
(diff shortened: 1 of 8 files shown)


src/view/actions/resetEventMergeId.jsx
```diff
@@ -36,10 +37,9 @@ const getSettings = ({ values }) => {
 };
 
 const validationSchema = object().shape({
-  eventMergeId: string().matches(
```
Aaronius, line 39: A value of an empty string wouldn't pass this `matches` test, but a valued of `undefined` will. Since `eventMergeId` is a required field, we need to use `.required` to ensure `undefined` fails validation.

src/view/actions/sendEvent.jsx
```diff
@@ -84,14 +85,8 @@ const getSettings = ({ values }) => {
 };
 
 const validationSchema = object().shape({
-  xdm: string().matches(
-    singleDataElementRegex,
-    "Please specify a data element"
```
Aaronius, line 89: I moved this text into a constant since we use it so much and so it will stay consistent.

jonsnyder (approved):

PR #143, opened 2021-06-03
Upgraded to react-spectrum 3. Added support for paging during schema selection

Aaronius:
We'll want to merge https://github.com/adobe/reactor-extension-alloy/pull/142 first and then merge this branch into the `react-spectrum` branch.

## Description
* Upgrades the XDM Object view to React-Spectrum 3. 
* Adds support for searching and paging when selecting a schema.
* Improves auto-selection of sandboxes and schemas.
* Improves error reporting, particularly for errors resulting from asynchronous operations.
* Improves handling of non-happy-path scenarios (e.g,, trying to save while a schema isn't selected or is still being loaded, race conditions with sandbox/schema selection)

I'd like to do some more manual testing once our org supports multiple sandboxes.

I apologize for the girth of this pull request. 😭 

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-1112
https://jira.corp.adobe.com/browse/PDCL-4567
https://jira.corp.adobe.com/browse/PDCL-6007

## Types of changes

## Checklist:


package.json
```diff
@@ -108,6 +110,6 @@
     "rollup-plugin-istanbul": "^3.0.0",
     "rollup-plugin-node-resolve": "^4.2.3",
     "stylus": "^0.54.8",
-    "testcafe": "^1.9.4"
+    "testcafe": "^1.14.2"
```
Aaronius, line 113: I upgraded testcafe primarily because it now provides a `t.scroll` and `t.scrollIntoView`, which I leveraged in my schema selection tests.

package.json
```diff
@@ -44,12 +44,14 @@
     "@adobe/alloy": "^2.4.0",
     "@adobe/react-spectrum": "^3.9.0",
     "@react-spectrum/combobox": "^3.0.0-beta.1",
+    "@react-stately/data": "^3.4.0",
     "@react/react-spectrum": "^2.34.0",
     "@spectrum-icons/illustrations": "^3.2.1",
     "@spectrum-icons/workflow": "^3.2.0",
-    "antd": "^4.12.3",
+    "antd": "^4.16.1",
```
Aaronius, line 51: I upgraded to the latest ant because I needed to do some work to exclude ant's global style rules, which were messing up React-Spectrum's styling of validation icons within textfields (at least). https://github.com/ant-design/ant-design/issues/9363

scripts/functional.js
```diff
@@ -60,7 +60,13 @@ const testsDir = src || path.join(__dirname, "../test/functional");
       return true;
     })
     .browsers("chrome")
-    .run();
+    .run({
+      // When using React's ErrorBoundary, even if the boundary catches an error, it doesn't swallow the
+      // error. Because the error isn't swallowed, TestCafe would typically immediately make the test fail,
+      // making it difficult to have a test that passes after asserting that the ErrorBoundary displayed
+      // a proper message.
+      skipJsErrors: true
```
Aaronius, line 68: This is an interesting situation. We're using our ErrorBoundary for displaying an error message not only for unforeseen JavaScript errors, but also for situations like when the user doesn't have appropriate AEP access. If we were _only_ using ErrorBoundary for unforeseen JavaScript errors (and handling foreseen errors in a different way), then we would probably leave `skipJsErrors` as `false`.

In our ErrorBoundary, we're also showing the message that comes from the error object. If the error is an unforeseen JavaScript error, we might not want to be showing that message to the user, since it's rather technical and the user can't do anything about it anyway. But when it's a foreseen error, like not having AEP access, the error object has a message that's user-friendly.

Anyway, this has me re-considering how we're dealing with error conditions, but I didn't really start thinking about it until this PR was mostly complete. I'd like to talk through it before changing it.

src/view/actions/sendEvent.jsx
```diff
@@ -121,7 +121,7 @@ const SendEvent = () => {
           <ExtensionViewForm
             initialValues={getInitialValues({ initInfo })}
             getSettings={getSettings}
-            validationSchema={validationSchema}
+            formikStateValidationSchema={validationSchema}
```
Aaronius, line 124: As you'll see, we now have `validateFormikState`, `formikStateValidationSchema`, and `validateNonFormikState` options.

src/view/components/alert.jsx
```diff
@@ -0,0 +1,70 @@
...
+    "notice",
+    "negative"
+  ]),
+  title: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
+  width: PropTypes.string,
+  className: PropTypes.string
+};
+
+export default Alert;
```
Aaronius, line 70: React-Spectrum doesn't have this available yet, so this is an interim Alert component that does a decent job of following the Spectrum spec.

src/view/components/extensionViewForm.jsx
```diff
@@ -7,17 +7,21 @@ import ExtensionViewContext from "./extensionViewContext";
...
+  validateNonFormikState,
   render
 }) => {
   const formikProps = useFormik({
     initialValues,
     enableReinitialize: true,
     onSubmit: () => {},
-    validate: wrapValidateWithErrorLogging(validate),
-    validationSchema,
-    validateOnChange: false
```
Aaronius, line 20: Notice that I removed `validateOnChange: false`, which means it will not validate on any form state change. I was playing around with disabling it in on the Send Event view, and it was okay there, but it made for some weird user experiences in the XDM object view (e.g., changing a selection within a radio group and the tree still showing that something was invalid until the radio group lost focus). By validating on every change, we're back in alignment with what we're doing on `main`.

src/view/components/extensionViewForm.jsx
```diff
@@ -7,17 +7,21 @@ import ExtensionViewContext from "./extensionViewContext";
...
+    validate: validateFormikState
+      ? wrapValidateWithErrorLogging(values => {
+          return validateFormikState({ values });
+        })
+      : undefined,
```
Aaronius, line 23: I'm still deciding on whether I'm hipster enough to use the `validateFormikState && wrapValidateWithErrorLogging(...)` pattern instead of the ternary.

src/view/components/extensionViewForm.jsx
```diff
@@ -51,11 +55,26 @@ const ExtensionViewForm = ({
...
+          // The docs say that the promise submitForm returns
+          // will be rejected if there are errors, but that is not the case.
+          // Therefore, after the promise is resolved, we pull formikProps.errors
+          // (which were set during submitForm()) to see if the form is valid.
+          // https://github.com/jaredpalmer/formik/issues/1580
+          formikPropsRef.current.setSubmitting(false);
+          return Object.keys(formikPropsRef.current.errors).length === 0;
+        });
+      const validateNonFormikStatePromise = Promise.resolve(
+        !validateNonFormikState || validateNonFormikState()
```
Aaronius, line 71: This is where we do view validation that's not related to Formik.

src/view/components/formikReactSpectrum3/checkbox.jsx
```diff
@@ -18,11 +18,13 @@ import FieldDescriptionAndError from "../fieldDescriptionAndError";
...
 const Checkbox = ({ name, description, width, ...otherProps }) => {
-  const [{ value }, { error }, { setValue, setTouched }] = useField(name);
+  const [{ value }, { touched, error }, { setValue, setTouched }] = useField(
+    name
+  );
   return (
     <FieldDescriptionAndError
       description={description}
-      error={error}
+      error={touched && error ? error : undefined}
```
Aaronius, line 27: I can't just do `touched && error` because when `touched` is `false`, it would feed a `false` value into `FieldDescriptionAndError`.

src/view/components/formikReactSpectrum3/checkbox.jsx
```diff
@@ -33,7 +35,7 @@ const Checkbox = ({ name, description, width, ...otherProps }) => {
         onBlur={() => {
           setTouched(true);
         }}
-        validationState={error ? "invalid" : ""}
+        validationState={touched && error ? "invalid" : undefined}
```
Aaronius, line 38: And I didn't really have to make this change, but I wanted it to be matchy matchy with `touched && error ? error : undefined`

src/view/configuration/configuration.jsx
```diff
@@ -441,19 +441,16 @@ const Configuration = ({
...
-      // a validation error, make sure the first accordion item containing
-      // an error is shown.
-      if (isSubmitting && !isValidating && errors && errors.instances) {
-        const instanceIndexContainingErrors = errors.instances.findIndex(
-          instance => instance
-        );
-        setSelectedAccordionIndex(instanceIndexContainingErrors);
-      }
-    },
-    formikProps
```
Aaronius, line 456: We no longer need to pass `formikProps` because we can use `useFormikContext` inside `useNewlyValidatedFormSubmission`.

src/view/dataElements/identityMap/fetchNamespaces.js
```diff
@@ -10,26 +10,21 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    });
+  } catch (e) {
+    // TODO: Should we be reporting an error instead?
+    return [];
+  }
 
-      return response.json();
-    })
-    .then(responseBody => responseBody)
-    .catch(() => []);
```
Aaronius, line 34: I've consolidated this logic in `fetchFromPlatform` (a part of the work I describe in https://jira.corp.adobe.com/browse/PDCL-4567).

src/view/dataElements/xdmObject.jsx
```diff
@@ -10,403 +10,252 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-}) => {
-  const { values: formState } = formikProps;
-  const [selectedNodeId, setSelectedNodeId] = useState();
+  // It might seem desirable to take advantage of the useImperativeHandle hook here,
+  // which does something very similar. Unfortunately, when using the useImperative
+  // handle, and the user adds an item to an array field (or possibly other things),
+  // React sets ref.current to undefined until this component finishes its next render.
+  // During that same period of time, Formik attempts to validate the form (because
+  // it validates on change) and the parent components calls childRef.current.validateFormikState
+  // which throws an error because React has momentarily set childRef.current to undefined.
```
Aaronius, line 84: This is weird, but I haven't really figured out a more appropriate way to deal with the situation.

src/view/dataElements/xdmObject.jsx
```diff
@@ -10,403 +10,252 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+            }}
+            render={({ initInfo, resetForm }) => {
+              return (
+                <XdmObject
+                  ref={childRef}
+                  initInfo={initInfo}
+                  resetForm={resetForm}
+                />
+              );
+            }}
```
Aaronius, line 258: I'd like to remove this `ref` stuff if there's a better way. Maybe it makes sense to not force the consumer of ExtensionViewForm to provide `initialValues`, `getSettings`, and validation methods up front, but instead provide a callback function through `render` or maybe use context to allow XDMObject to register its methods?

src/view/dataElements/xdmObject/components/booleanEdit.jsx
```diff
@@ -75,96 +44,67 @@ const BooleanInput = props => {
...
+        >
+          <Radio data-test-id="constantNoValueField" value="">
+            No Value
+          </Radio>
+          <Radio data-test-id="constantTrueField" value>
+            True
+          </Radio>
+          <Radio data-test-id="constantFalseField" value={false}>
+            False
+          </Radio>
```
Aaronius, line 95: I switched from a Select to Radios because (1) it's what we should be using according to [spectrum guidelines](https://spectrum.adobe.com/page/combo-box/#Combo-box,-picker,-or-radio-buttons?) (2) Select/Picker doesn't support boolean values (mostly because they're used for React keys which must be strings) so you have to use strings instead and have a translation layer to get a boolean into formik state (see `convertActualValueToBooleanOptionValue` and `convertBooleanOptionValueToActualValue` in the old code for what that looks like).

src/view/dataElements/xdmObject/components/noSelectedNodeView.jsx
```diff
@@ -47,50 +44,39 @@ const NoSelectedNodeView = props => {
...
+          >
+            A full circle indicates all of the attributes have been populated.
+          </IndicatorDescription>
+          <IndicatorDescription indicator={<AsteriskIcon size="XS" />}>
+            Fields that may be auto-populated when this data element is passed
+            to the XDM option of the &quot;Send Event&quot; action have this
+            icon. Hovering over the icon shows a popup explaining when the field
+            will be auto-populated.
+          </IndicatorDescription>
+        </Flex>
```
Aaronius, line 72: We could probably avoid a lot of our utility classes by using React-Spectrum's Flex and View components. I'm still feeling that out.

src/view/dataElements/xdmObject/components/nodeEdit.jsx
```diff
@@ -23,28 +23,29 @@ import {
...
-import integerEdit from "./integerEdit";
-import numberEdit from "./numberEdit";
-import objectEdit from "./objectEdit";
-import stringEdit from "./stringEdit";
+import ArrayEdit from "./arrayEdit";
+import BooleanEdit from "./booleanEdit";
+import IntegerEdit from "./integerEdit";
+import NumberEdit from "./numberEdit";
+import ObjectEdit from "./objectEdit";
+import StringEdit from "./stringEdit";
```
Aaronius, line 31: Not really necessary but I noticed they weren't capitalized and it bugged me since they're components.

src/view/dataElements/xdmObject/components/sandboxSelector.jsx
```diff
@@ -0,0 +1,126 @@
...
+      try {
+        ({ results: sandboxes } = await fetchSandboxes({
+          orgId,
+          imsAccess,
+          signal
+        }));
+      } catch (e) {
+        if (e.name !== "AbortError") {
+          console.error(e);
+          reportAsyncError(new Error(`Failed to load sandboxes. ${e.message}`));
```
Aaronius, line 57: I did consider moving console.error into `reportAsyncError` and having `reportAsyncError` also accept something like `errorMessage` (which would be `Failed to load sandboxes` in this case) and `originatingError` (which would be `e` in this case), but it felt like we were overloading `reportAsyncError`. We could make an abstraction on top of `reportAsyncError`, but I was a bit ambivalent. You'll also notice in other parts of the code that we don't always have an "originating error".

src/view/dataElements/xdmObject/helpers/fetchSchemasMeta.js
```diff
@@ -10,56 +10,60 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
   const params = new URLSearchParams();
   params.append("orderby", "title");
   params.append("property", `meta:extends==${metaExtends}`);
 
   if (search) {
-    params.append("property", `title~${search}`);
+    // We escape regex special characters because ~ in the querystring
+    // searches by regex and we don't want to search by regex.
+    params.append("property", `title~${escapeStringRegexp(search)}`);
```
Aaronius, line 38: `escapeStringRegexp` fixes a bug I noticed.

src/view/dataElements/xdmObject/helpers/getInitialFormState.js
```diff
@@ -85,7 +85,6 @@ const getInitialFormStateNode = ({ sandboxMeta, schema, value, nodePath }) => {
     // from their parent arrays. We want the ID to remain constant
     // for as long as the node exists.
     id: generateNodeId(),
-    sandboxMeta,
```
Aaronius, line 88: `sandboxMeta` was a relic no longer being used.

src/view/dataElements/xdmObject/helpers/mainViewState.js
```diff
@@ -0,0 +1,79 @@
...
+
+export const ACTION_TYPES = {
+  DEFAULT_SANDBOX_AND_SCHEMA_LOADED: "defaultSandboxAndSchemaLoaded",
+  SELECTED_SANDBOX_CHANGED: "selectedSandboxChanged",
+  SCHEMA_LOADED: "schemaLoaded",
+  SELECTED_SCHEMA_META_CHANGED: "selectedSchemaMetaChanged",
+  SELECTED_NODE_ID_CHANGED: "selectedNodeIdChanged",
+  UPDATE_SHOW_EDITOR_NOT_READY_VALIDATION_ERROR:
+    "updateShowEditorNotReadyValidationError"
+};
```
Aaronius, line 23: I've [debated proper capitalization methods](https://adobedx.slack.com/archives/GPKKPT0Q1/p1622237500146300). It seems we've been inconsistent in these cases where we basically have an enum object. Airbnb [has a stance](https://github.com/airbnb/javascript/#naming--uppercase) (`ACTION_TYPES` uppercase, `defaultSandboxAndSchemaLoaded` lowercase), which I'd be cool with, but it gets interesting when we have this: `import * as STATUS from "../constants/mainViewStatus";` The constants are named exports inside `mainViewStatus.js`, so they're uppercased...meaning we end up with `STATUS.SOMETHING` rather than `STATUS.something`.

src/view/dataElements/xdmObject/helpers/platform.js
```diff
@@ -1,90 +0,0 @@
...
-          throw new Error(
-            "Your user account is not enabled for AEP access. Please contact your organization administrator."
-          );
-        }
-      }
-
-      return parsedResponse;
-    });
-  }
-};
```
Aaronius, line 90: Most of this has moved to `fetchFromPlatform.js`.

src/view/utils/useExtensionBridge.js
```diff
@@ -12,8 +12,8 @@ governing permissions and limitations under the License.
 import { useRef, useEffect } from "react";
 
 export default currentViewMethods => {
-  const viewMethods = useRef();
-  viewMethods.current = currentViewMethods;
+  const viewMethodsRef = useRef();
```
Aaronius, line 15: I always forget I'm dealing with a ref, so I'm trying to suffix all ref variables with `Ref`.

test/functional/dataElements/xdmObject/helpers/xdmTree.js
```diff
@@ -19,63 +19,63 @@ import {
...
-        return t.expect(populationIndicator.hasClass("is-empty")).ok();
+      expectExists: async () => {
+        await switchToIframe();
+        await t.expect(node.exists).ok();
       },
-      expectBlank() {
-        return t.expect(populationIndicator.exists).notOk();
+      expectNotExists: async () => {
+        await switchToIframe();
+        await t.expect(node.exists).notOk();
```
Aaronius, line 77: There's some context as to why I changed the organization of this helper. I wanted to have an `expectExists` method that would tell me if the tree itself actually existed, rather than a node on the tree. It seemed appropriate, then, not put all the node-based stuff under a `node` method. I didn't actually end up adding an `expectExists` method for the tree itself, but instead added an `editor.js` helper which has an `expectExists` that tells me whether the editor exists (the editor contains the tree), which fulfilled my needs. I left these changes in the tree though because I feel it is a better structure. Unfortunately, I had to change a bunch of the usages of this helper, which makes of a larger PR diff. Sorry.

test/functional/dataElements/xdmObject/xdmObjectEditing.spec.js
```diff
@@ -0,0 +1,454 @@
...
+  await xdmTree.node("_id").click();
+  await stringEdit.expectNotExists();
+});
+
+test("allows you to edit context fields", async () => {
+  await initializeExtensionView();
+  await xdmTree.node("environment").toggleExpansion();
+  await xdmTree.node("type").click();
+  await stringEdit.expectExists();
+});
```
Aaronius, line 454: This files was just a rename from xdmObject.spec.js. As far as I recall, there were no significant changes--mostly just changes to reflect the new xdmTree helper structure.

test/functional/dataElements/xdmObject/xdmObjectFieldValidation.spec.js
```diff
@@ -10,72 +10,58 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 // disablePageReloads is not a publicized feature, but it sure helps speed up tests.
 // https://github.com/DevExpress/testcafe/issues/1770
-fixture("XDM Object Population Indicator")
+fixture("XDM Object Validation")
   .disablePageReloads.page("http://localhost:3000/viewSandbox.html")
   .meta("requiresAdobeIOIntegration", true);
 
 test("arrays with no values are invalid", async () => {
-  await initializeExtensionView();
-  await selectSchemaFromSchemasMeta();
```
Aaronius, line 38: If/When we add more schemas to our sandbox, we might need to add something like `selectSchemaFromSchemasMeta` back in, but right now we only have one schema in the sandbox so it gets automatically selected.

jonsnyder (approved):
Great work! I think using the reducer helps to organize the logic really well. "usePagedComboBox" came out really well and easy to follow. I think that simplifies things over using asyncList.

I tested the UI locally. Here are two issues I found:
1. The font on the new XDM tree is slightly larger. It may make sense to widen the tree section to avoid cutting off long names in deeply nested objects. Ideally it would grow based on the tree expansion size.
2. In a deeply nested array, "Item" and "1" are on different lines. (i.e. _experience > decisioning > Item 1 > scopeDetails > interationMeasurement > Item 1) Maybe we should use a non-breaking space character between Item and 1.

scripts/functional.js
Aaronius: Yeah, I agree. I'm going to see how some of this error management plays out in the other views and decide if/when we want to change how we're handling foreseen error situations.

src/view/components/extensionViewForm.jsx
Aaronius: Sorry, I meant:

> which means it will **now** validate on any form state change

but you probably understood that.

src/view/dataElements/xdmObject/helpers/mainViewState.js
Aaronius: I think I like that option less than just keeping them all caps.

src/view/dataElements/xdmObject.jsx
Aaronius: It's kind of a chicken or the egg problem. If I render ExtensionViewForm inside the XDMObject, then the XDM Object won't have immediate access to `resetForm` which it uses in `useOnSandboxSelectionChange`, `useOnSchemaMetaSelectionChange`, and within a `useEffect`. I'll show you, and then show you the new API I've been playing with.

Aaronius:
Thanks for the review, @jonsnyder. I believe I've addressed your feedback.

Aaronius:
I'm going to merge this to help avoid conflicts. I know you were interested in testing out the view when multiple schemas are available (and I wanted to test it out when multiple sandboxes are available). I figure we can do that on the react-spectrum branch and go from there.

PR #145, opened 2021-06-10
PDCL-6013 Updated Identity Map to react-spectrum 3 using refactored ExtensionViewForm.

Aaronius:
…

## Description
Updated Identity Map to react-spectrum 3 using refactored ExtensionViewForm. @jonsnyder, after our discussion about ExtensionViewForm, I full expect some of that to change. We'll likely have conflicts, but I can help sort that out.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-6013

## Types of changes

## Checklist:


src/view/actions/sendEvent.jsx
```diff
@@ -113,101 +114,123 @@ const knownEventTypeOptions = [
...
+            }}
           />
         );
       }}
     />
   );
 };
 
-render(SendEvent);
+render(SendEventExtensionView);
```
Aaronius, line 236: The changes to SendEvent were just to accommodate the refactored ExtensionViewForm.

src/view/components/decisionScopes.jsx
```diff
@@ -77,104 +77,122 @@ const validationSchema = object().shape({
...
+    <ExtensionViewForm
+      render={props => {
+        return <DecisionScopes {...props} />;
       }}
     />
   );
 };
 
-export default DecisionScopes;
+export default DecisionScopesExtensionView;
```
Aaronius, line 198: The changes to DecisionScopes were just to accommodate the refactored ExtensionViewForm.

src/view/components/extensionViewForm.jsx
```diff
@@ -1,76 +1,96 @@
...
-  initialValues,
-  getSettings,
-  validateFormikState = () => undefined,
-  formikStateValidationSchema,
-  validateNonFormikState = () => true,
-  render
-}) => {
-  const formikProps = useFormik({
-    initialValues,
-    enableReinitialize: true,
```
Aaronius, line 17: I had misunderstood what `enableReinitialize` was for when I had set it to `true`. I thought it meant that it would allow us to re-initialize the form using `resetForm`, but it actually means that if `initialValues` changes, that the form will automatically reset. In our case, whenever we reset the form, we explicitly use `resetForm`. What I discovered is that there's a bug in Formik when we call `resetForm` on our first render (as now happens on SendEvent, for example). Formik resets its values to whatever we pass into `resetForm`, but then it resets its values again back to what `initialValues` was, which is `undefined` in our case (related to https://github.com/formium/formik/issues/2440). Fortunately, we don't need/want `enableReinitialize: true`, so removing it also avoids the bug.

src/view/components/extensionViewForm.jsx
```diff
@@ -1,76 +1,96 @@
...
+      // to handle them in a proper manner.
+      try {
+        if (viewRegistrationRef.current?.validateFormikState) {
+          errors = viewRegistrationRef.current.validateFormikState({ values });
+        }
+      } catch (error) {
+        reportAsyncError(
+          new Error("An error occurred while validating the view.")
+        );
+      }
```
Aaronius, line 28: I moved this logic out of `wrapValidateWithErrorLogging`.

src/view/components/identityWrapper.jsx
```diff
@@ -1,215 +0,0 @@
...
-      />
-    </React.Fragment>
-  );
-}
-
-IdentityWrapper.propTypes = {
-  values: PropTypes.object
-};
-
-export default IdentityWrapper;
```
Aaronius, line 215: I moved this into identityMap.jsx to be more consistent with our other views.

src/view/components/namespaceComponent.jsx
```diff
@@ -1,61 +1,61 @@
...
 import React from "react";
-import Textfield from "@react/react-spectrum/Textfield";
 import PropTypes from "prop-types";
-import WrappedField from "./wrappedField";
+import { Item } from "@adobe/react-spectrum";
+import { Picker, TextField } from "./formikReactSpectrum3";
 
-const getSelectedNamespace = (options, namespace) => {
-  if (options.length < 1) {
+const getSelectedNamespace = (namespaces, selectedNamespaceCode) => {
```
Aaronius, line 6: We have several references to `namespace`. Sometimes it referred to the namespace object, while sometimes it referred specifically to the namespace code that's on a namespace object. I did some renaming to disambiguate.

src/view/configuration/configuration.jsx
```diff
@@ -441,11 +434,11 @@ const Configuration = ({
     };
   }, [firstInstance.edgeConfigId, firstInstance.edgeConfigInputMethod]);
 
-  useNewlyValidatedFormSubmission(() => {
+  useNewlyValidatedFormSubmission(errors => {
     // If the user just tried to save the configuration and there's
     // a validation error, make sure the first accordion item containing
     // an error is shown.
-    if (isSubmitting && !isValidating && errors && errors.instances) {
+    if (errors && errors.instances) {
```
Aaronius, line 441: I improved `useNewlyValidatedFormSubmission` so that some of this logic became unnecessary.

src/view/configuration/configuration.jsx
```diff
@@ -1030,40 +1023,35 @@ const Configuration = ({
...
-                            icon={<Delete />}
-                            variant="action"
-                            disabled={values.instances.length === 1}
-                          />
-                          {values.instances.length === 1 ? (
-                            <span className="Note u-gapLeft">
-                              You must have at least one instance to use this
-                              extension.
-                            </span>
-                          ) : null}
```
Aaronius, line 1047: Previously we were showing the Delete Instance button but if there was only one instance we would disable it and add this note next to it. I don't think it really improved the user experience to show and disable it and it cluttered the UI for the majority of cases where only one instance is configured. This was especially the case for IdentityMap, where we have identities and then identifiers nested inside each identity. It felt like button overload.

src/view/dataElements/xdmObject.jsx
```diff
@@ -226,39 +227,22 @@ const XdmObject = forwardRef(({ initInfo, resetForm }, ref) => {
...
-            render={({ initInfo, resetForm }) => {
-              return (
-                <XdmObject
-                  ref={childRef}
-                  initInfo={initInfo}
-                  resetForm={resetForm}
-                />
-              );
+            render={props => {
+              return <XdmObject {...props} />;
```
Aaronius, line 245: The changes to XdmObject were just to accommodate the refactored ExtensionViewForm.

src/view/components/namespaceComponent.jsx
```diff
@@ -1,61 +1,61 @@
...
+  return selectedNamespace ? (
+    <Picker
+      data-test-id={`namespacePicker${index}Field`}
+      label="Namespace"
+      name={name}
+      items={namespaces}
+      width="size-5000"
+      isRequired
+    >
+      {namespace => <Item key={namespace.code}>{namespace.name}</Item>}
```
Aaronius, line 41: We're now displaying the `name` of the namespace (which seems more user-friendly) rather than the `code`.

jfkhoury: Nice!

jfkhoury (approved):
Looks good, considering I haven't done UI work in a while :)

PR #147, opened 2021-06-16
Segregated "user-reportable errors" from errors whose messages really…

Aaronius:
… shouldn't be exposed to customers.

## Description
When an error occurs, sometimes we catch that error and throw a different error with a message that's appropriately customer-facing (e.g., `Your organization is not provisioned for Adobe Experience Platform. Please contact your organization administrator.`). Unless our code is 100% bulletproof, there are likely other errors that are thrown as well that we don't catch and throw a new error for (e.g., `ReferenceError: nonexistent is not defined`). The messages from these latter errors were previously being displayed to the user. We don't want to be exposing these types of error messages to customers though because they're daunting and the user can't do anything about them anyway.

This PR segregates the two types of errors by providing a UserReportableError class that we can use to throw the former types of errors. When the error boundary sees an error that is not a UserReportableError, it will just display `An unexpected error occurred. Please try again later.` rather than the error's actual message.

Also note that TestCafe by default fails a test if it sees there are any unhandled errors that occur during the test. Because our error boundary doesn't prevent errors from bubbling up the stack (they aren't "caught" in the literal sense), TestCafe would see them as unhandled and fail the test, even if we were trying to test an error case (e.g., that we were telling the user that they didn't have AEP access when they didn't). To get around this issue, we had configured TestCafe with `ignoreJsErrors: true`. Unforunately, this made it so that _unforseen_ errors were not causing tests to fail, when really it would be great if those types of errors failed the test. This PR makes it so that UserReportableError errors are not considered unhandled, so only the errors that are not UserReportableError instances will be seen as unhandled by TestCafe and fail the tests. This allows up to remove the `ignoreJsErrors: true` TestCafe configuration.

Clear as mud?

## Motivation and Context
Better error reporting for our customers.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/xdmObject/helpers/schemaSelection/loadDefaultSandbox.js
+++ b/src/view/dataElements/xdmObject/helpers/schemaSelection/loadDefaultSandbox.js
@@ -12,6 +12,7 @@ governing permissions and limitations under the License.
 
 import fetchSandboxes from "../fetchSandboxes";
 import DEFAULT_SANDBOX_NAME from "../../constants/defaultSandboxName";
+import UserReportableError from "../../../../errors/userReportableError";
 
 const loadDefaultSandbox = async ({
   orgId,
@@ -27,15 +28,21 @@ const loadDefaultSandbox = async ({
       imsAccess
     }));
   } catch (e) {
-    console.error(e);
-    reportAsyncError(new Error(`Failed to load sandboxes. ${e.message}`));
+    reportAsyncError(
+      new UserReportableError("Failed to load sandboxes.", {
+        originatingError: e
+      })
+    );
     return undefined;
   }
 
   if (!sandboxes.length) {
     reportAsyncError(
-      new Error(
-        "You do not have access to any sandboxes. Please contact your administrator to be assigned appropriate rights."
+      new UserReportableError(
+        "You do not have access to any sandboxes. Please contact your administrator to be assigned appropriate rights.",
+        {
+          additionalInfoUrl: "https://adobe.ly/3gHkqLF"
+        }
       )
     );
     return undefined;
@@ -60,7 +67,7 @@ const loadDefaultSandbox = async ({
 
     if (!defaultSandbox) {
       reportAsyncError(
-        new Error(
+        new UserReportableError(
           "The sandbox used to build the XDM object no longer exists. You will need to re-create this data element using a schema from a different sandbox."
         )
       );
--- a/src/view/utils/fetchFromPlatform.js
+++ b/src/view/utils/fetchFromPlatform.js
@@ -23,11 +23,14 @@ const IMS_HOST_PREFIX_STAGING = "ims-na1-stg1";
 const ERROR_CODE_INVALID_ACCESS_TOKEN = "401013";
 const ERROR_CODE_USER_REGION_MISSING = "403027";
 
+// The internet connection may not actually be the problem here. There could
+// be other problems with the SSL handshake, CORS preflight request, etc.
 const ERROR_UNABLE_TO_CONNECT_TO_SERVER =
-  "A connection to the server could not be established.";
+  "A connection to the server could not be established. You may be disconnected from the internet. Please check your connection and try again.";
 const ERROR_UNEXPECTED_SERVER_RESPONSE =
   "An unexpected server response was received.";
-const ERROR_INVALID_ACCESS_TOKEN = "Your access token appears to be invalid.";
+const ERROR_INVALID_ACCESS_TOKEN =
+  "Your access token appears to be invalid. Please try logging in again.";
 const ERROR_NO_AEP_ACCESS =
   "Your organization is not provisioned for Adobe Experience Platform. Please contact your organization administrator.";
 const ERROR_RESOURCE_NOT_FOUND = "The resource was not found.";
--- a/src/view/utils/fetchFromPlatform.js
+++ b/src/view/utils/fetchFromPlatform.js
@@ -12,6 +12,7 @@ governing permissions and limitations under the License.
 
 import getBaseRequestHeaders from "./getBaseRequestHeaders";
 import * as HTTP_STATUS from "../constants/httpStatus";
+import UserReportableError from "../errors/userReportableError";
 
 const PLATFORM_HOST_PROD = "https://platform.adobe.io";
 const PLATFORM_HOST_STAGING = "https://platform-stage.adobe.io";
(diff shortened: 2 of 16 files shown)


src/view/dataElements/xdmObject/helpers/schemaSelection/loadDefaultSandbox.js
```diff
@@ -27,15 +28,21 @@ const loadDefaultSandbox = async ({
...
   }
 
   if (!sandboxes.length) {
     reportAsyncError(
-      new Error(
-        "You do not have access to any sandboxes. Please contact your administrator to be assigned appropriate rights."
+      new UserReportableError(
+        "You do not have access to any sandboxes. Please contact your administrator to be assigned appropriate rights.",
+        {
+          additionalInfoUrl: "https://adobe.ly/3gHkqLF"
```
Aaronius, line 44: @jfkhoury was thinking a link would be nice for this error, so this is how it's done.

src/view/utils/fetchFromPlatform.js
```diff
@@ -22,11 +23,14 @@ const IMS_HOST_PREFIX_STAGING = "ims-na1-stg1";
...
-const ERROR_UNEXPECTED_SERVER_RESPONSE = "Unexpected server response.";
+const ERROR_UNABLE_TO_CONNECT_TO_SERVER =
+  "A connection to the server could not be established.";
+const ERROR_UNEXPECTED_SERVER_RESPONSE =
+  "An unexpected server response was received.";
 const ERROR_INVALID_ACCESS_TOKEN = "Your access token appears to be invalid.";
 const ERROR_NO_AEP_ACCESS =
-  "Your user account is not enabled for AEP access. Please contact your organization administrator.";
-const ERROR_RESOURCE_NOT_FOUND = "Resource not found.";
+  "Your organization is not provisioned for Adobe Experience Platform. Please contact your organization administrator.";
```
Aaronius, line 32: As far as I can tell from Slack conversations, this scenario ("user region is missing") is because the org isn't properly provisioned and not so much of a user account thing.

test/functional/dataElements/xdmObject/helpers/platformMocks.js
```diff
@@ -314,9 +308,6 @@ export const schemasMetaPagingTitles = [
 export const schemasMetaPagingMock = RequestMock()
   .onRequestTo({
     url: SCHEMAS_ENDPOINT_REGEX,
-    headers: {
-      "x-sandbox-name": "prod"
-    },
```
Aaronius, line 319: This wasn't actually doing anything. `headers` isn't a valid option for `onRequestTo`.

src/view/utils/fetchFromPlatform.js
```diff
@@ -63,17 +67,25 @@ export default async ({ orgId, imsAccess, path, params, headers, signal }) => {
...
+        ...baseRequestHeaders,
+        ...headers
+      },
+      signal
+    });
+  } catch (e) {
+    if (e.name === "AbortError") {
+      throw e;
+    }
+    throw new UserReportableError(ERROR_UNABLE_TO_CONNECT_TO_SERVER);
```
Aaronius, line 84: I added this in case the user didn't have an internet connection or the CORS request failed or something like that.

jonsnyder (approved):
This is definitely an improvement.

There is one case that isn't covered here, but I'm not sure if it could be solved. What happens if a UserReportableError is thrown, but it is not caught by reportAsyncError? i.e. it was thrown in an async function, but never reported to the error boundary. Should that fail automated tests?

src/view/dataElements/xdmObject/helpers/schemaSelection/loadDefaultSandbox.js
```diff
@@ -27,15 +28,21 @@ const loadDefaultSandbox = async ({
...
+    );
     return undefined;
   }
 
   if (!sandboxes.length) {
     reportAsyncError(
-      new Error(
-        "You do not have access to any sandboxes. Please contact your administrator to be assigned appropriate rights."
+      new UserReportableError(
+        "You do not have access to any sandboxes. Please contact your administrator to be assigned appropriate rights.",
```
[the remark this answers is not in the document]
Aaronius: I think I can add some actionability on a couple of the others. Most errors though, particularly the non-user-reportable errors, don't seem to have anything user-actionable. For example, if we have some sort of unforeseen JavaScript error or the server blew up and is spewing 500 status codes, I don't think there's anything a customer can do to fix the problem. I don't have confidence that refreshing the Launch page or clicking buttons will help, so "try again later" is just that--try again sometime later because hopefully by that point we've fixed the problem on our side. The one actionable thing they could probably do is report the issue through their customer rep. Do you have any opinion on what it should say in these scenarios?

Aaronius:
> This is definitely an improvement.
> 
> There is one case that isn't covered here, but I'm not sure if it could be solved. What happens if a UserReportableError is thrown, but it is not caught by reportAsyncError? i.e. it was thrown in an async function, but never reported to the error boundary. Should that fail automated tests?

Yeah, that's a good point. It probably should fail the automated test. I'll add a commit that makes this happen.

src/view/dataElements/xdmObject/helpers/schemaSelection/loadDefaultSandbox.js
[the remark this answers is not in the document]
Aaronius: BTW, I logged this issue so that we're more proactive at detecting and fixing errors occurring for customers: https://jira.corp.adobe.com/browse/PDCL-6348

Aaronius:
> > This is definitely an improvement.
> > There is one case that isn't covered here, but I'm not sure if it could be solved. What happens if a UserReportableError is thrown, but it is not caught by reportAsyncError? i.e. it was thrown in an async function, but never reported to the error boundary. Should that fail automated tests?
> 
> Yeah, that's a good point. It probably should fail the automated test. I'll add a commit that makes this happen.

I don't think this is feasible. I was planning on adding an `isReportedToUser` property on the error object that the error boundary could set to `true`, then the global error event listener could check that boolean to see if it had been reported. Unfortunately, React's error boundary only exposes the error object to our code _after_ the global error event listener has received it. I've updated my code comment to provide more detail about this.

Aaronius:
@jonsnyder I'm going to merge this, but I'd still like to hear your opinion on actionable messaging.

src/view/dataElements/xdmObject/helpers/schemaSelection/loadDefaultSandbox.js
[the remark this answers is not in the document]
jonsnyder: I like the ticket you created. That is much better than giving the user a link to report the error. I'm glad you looked through the errors.

PR #148, opened 2021-06-21
Convert setConsent action to React-spectrum 3. Only have one Formik on a page.

jonsnyder:
## Types of changes

## Checklist:


src/view/utils/nameUtils.js
```diff
@@ -0,0 +1,115 @@
...
+
+This is the parsing grammar used:
+
+name -> id nameTail
+nameTail -> "[" number "]" nameTail
+            "." name
+            end
+id -> /[^\[\]\.]+/
+number -> /[0-9]+/
+*/
```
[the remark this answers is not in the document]
Aaronius: I'm impressed by the depths you went on this. It looks very similar to https://www.npmjs.com/package/object-path. Do you think it provides additional value or can we offload the work to that package?

src/view/utils/usePartialForm.js
```diff
@@ -0,0 +1,43 @@
...
+  configRef.current.validateFormikState = validateFormikState;
+  configRef.current.formikStateValidationSchema = formikStateValidationSchema;
+  configRef.current.validateNonFormikState = validateNonFormikState;
+  configRef.current.initialized = initialized;
+  configRef.current.initializedName = initializedName;
+
+  useEffect(() => {
+    registerPartialForm(configRef.current);
+    return () => {
+      dropPartialForm(configRef.current);
```
Aaronius, line 34: We discussed this separately, but just as a note: partialForm will need to be adjusted in order for it to be usable within components that dismount but should remain part of the state (e.g., tab panels).

src/view/components/consentObjects.jsx
```diff
@@ -0,0 +1,253 @@
...
+                                  label="GDPR Contains Personal Data"
+                                  name={`consent[${index}].gdprContainsPersonalData`}
+                                  dataTestIdPrefix="gdprContainsPersonalData"
+                                  options={[YES, NO]}
+                                  defaultValue={NO.value}
+                                />
+                              </>
+                            )}
+                            {consent.length > 1 && (
+                              <div className="u-gapTop">
```
Aaronius, line 224: Upon further review (we discussed this previously), this `<div>` isn't necessary in order to keep the button from being expanded to 100% width. Without the `div`, the button would typically be expanded to 100% width because it's within a `<Flex direction="column">` and by default flex items are stretched along the cross axis of the flex parent. There are a few ways to solve this:

* Set `alignItems="start"` on the Flex element.
* Set `alignSelf="flex-start"` on the Button element.
* Do what you did here.

src/view/components/dataElementRadioChoice.jsx
```diff
@@ -0,0 +1,132 @@
+import { string, object } from "yup";
+import React from "react";
+import { Radio } from "@adobe/react-spectrum";
+import PropTypes from "prop-types";
+import { useField } from "formik";
+import singleDataElementRegex from "../constants/singleDataElementRegex";
+import DataElementSelector from "./dataElementSelector";
+import ImperativeForm from "./partialForm";
```
Aaronius, line 8: I think the name here should be `PartialForm`.

Aaronius (commented):
I plan on looking through this some more tomorrow, but I'm posting the comments I have so far.

Aaronius:
I think exporting getInitialValues, getSettings, and validate/validationSchema (as named exports) from each component and then composing them together at the top-level component would be preferable over having PartialForms. This is similar to what we were doing with decision scopes in Send Event before switching to ExtensionViewForm. It’s definitely not as encapsulated as using these nested PartialForms, but it still allows the extension-bridge-related logic to primarily reside next to the component code while not requiring the component to be rendered in order for its logic to be effective. I’m mostly worried about how we would need to deal with non-rendered components when using PartialForm and the complexity it will bring. This will be an immediate need because we'll want to use OptionsWithDataElement in Configuration, but OptionsWithDataElement will only be rendered for the active tab. If there's a simple way to deal with that situation while still using PartialForms, I'd be interested in learning more.

src/view/components/spectrum3ExtensionView.jsx
```diff
@@ -1,78 +1,193 @@
...
+    return null;
+  }
 
   return (
-    <ExtensionViewContext.Provider value={context}>
-      {render({ initInfo })}
-    </ExtensionViewContext.Provider>
+    <>
+      <ExtensionViewContext.Provider value={context}>
+        <Form>
```
Aaronius, line 176: I think putting `Form` here might be presumptuous. I think the only real reason we're using `Form` is because it provides a bit of styling on fields. One of the styles (top-margin) is provided through this declaration block:

```
._spectrum-Form_36b9b._spectrum-Form--positionTop_36b9b>* {
    margin-top: var(--spectrum-global-dimension-size-100);
    width: 100%
}
```

Notice that the fields receiving `margin-top` will be the direct children of the form. Considering the `Form` placed here will likely have children that are not the fields (for example, it might be Tabs), I don't think the elements that will be receiving the top margin are the ones we want to be receiving the top margin.

I'm still trying to figure out how much value we're getting out of Form and how to use it most appropriately.

src/view/components/consentObjects.jsx
jonsnyder: Thanks for the help. I need to study Flex alignment at some point. When I learned CSS, I still had to support IE 6.

jonsnyder:
Aaron and I talked it over, and decided not to go with the approach in this PR, but just have one ImperativeForm in each view for the reasons he outlined above.

jfkhoury:
Thank you both!

PR #152, opened 2021-06-24
Update setConsent to react-spectrum3, fixes to RadioGroupWithDataElement

jonsnyder:
## Types of changes

## Checklist:


src/view/components/rawDataElementSelector.styl
```diff
@@ -11,5 +11,5 @@ governing permissions and limitations under the License.
 */
 
 .DataElementSelector-buttonAdjustedForLabel {
-  margin-top: 24px;
+  margin-top: 24px !important;
```
jonsnyder, line 14: I don't understand why this is necessary.

src/view/components/formikReactSpectrum3/radioGroupWithDataElement.jsx
```diff
@@ -75,14 +75,17 @@ const RadioGroupWithDataElement = ({
   });
```
jonsnyder, line 75: The setConsent tests found some bugs in the radioGroupWithDataElement component.

Aaronius:
If I load the view using this for settings:

```
{
    "instanceName": "alloy",
    "identityMap": "%dataElement9865%",
    "consent": "%dataElement2sdf861%"
  }
```

and then switch from "use a data element" to "fill out a form", I get an error caught by the error boundary.

image

Aaronius:
When I load the view using this for settings:

```
{
  "instanceName": "alloy",
  "consent": [
    {
      "standard": "Adobe",
      "version": "1.0",
      "value": {
        "general": "%dataElement9680%"
      }
    },
    {
      "standard": "Adobe",
      "version": "2.0",
      "value": "%dataElement8346%"
    },
    {
      "standard": "Adobe",
      "version": "1.0",
      "value": {
        "general": "in"
      }
    }
  ]
}
```

and I switch the first consent from "use a data element" to "in" and then back to "use a data element", the value that was in "use a data element" is gone, which I believe we were trying to avoid.

src/view/components/rawDataElementSelector.styl
Aaronius: I dug into this. It turns out the styles from `@adobe/react-spectrum` are bundled in our CSS file before `.DataElementSelector-buttonAdjustedForLabel`, which is good, but the styles from `@react-spectrum/combobox` are bundled into the CSS file _after_ `.DataElementSelector-buttonAdjustedForLabel`. The combobox styles include a `._spectrum-ActionButton_7a745,._spectrum-Button_7a745,._spectrum-ClearButton_7a745,._spectrum-FieldButton_7a745,._spectrum-LogicButton_7a745` selector that sets margin to 0. Since this selector comes after `.DataElementSelector-buttonAdjustedForLabel`, its 0 margin wins.

I don't think there's a great way of forcing the ComboBox's styles to be bundled into the CSS file before our own styles (I've asked on the react-spectrum channel. I'll let you know if I hear otherwise.). Considering this, using `!important` here is a reasonable workaround, but I think it would be better to add a `marginTop="size-300"` prop to the button inside `rawDataElementSelector.jsx` instead of having this class. I probably should have done that in the first place.

jonsnyder:
@Aaronius thank you for taking the time to test it. I've fixed that error and added an automated test for it.

src/view/components/formikReactSpectrum3/radioGroupWithDataElement.jsx
```diff
@@ -77,18 +77,26 @@ const RadioGroupWithDataElement = ({
...
+  ) {
+    radioValue = value;
+    dataElementText = dataElementValue.dataElement;
   } else {
     radioValue = "dataElement";
     dataElementText = value;
   }
 
   const setValues = (newRadioValue, newDataElement) => {
+    console.log("setValues", newRadioValue, newDataElement);
```
Aaronius, line 99: Remove.

Aaronius (approved):

jonsnyder:
@Aaronius Thanks for finding these two bugs!  I fixed them and addressed your comments. I'm not sure about the spacing issue. Let's meet on Monday to discuss the spacing.

src/view/components/rawDataElementSelector.jsx
```diff
@@ -45,9 +44,7 @@ const RawDataElementSelector = ({
         isQuiet
         onPress={openDataElementSelector}
         aria-label="Select data element"
-        UNSAFE_className={
-          adjustForLabel ? "DataElementSelector-buttonAdjustedForLabel" : ""
-        }
+        UNSAFE_style={adjustForLabel ? { marginTop: "size-300" } : {}}
```
Aaronius, line 47: This won't work because `size-300` isn't a valid CSS value. You'll need to use the `marginTop` prop. It's a prop provided by React-Spectrum that does the interpretation from `size-300` to 24px.

PR #153, opened 2021-06-28
Make spacing between elements consistent

jonsnyder:
## Description

We have been using Form to get the margin between elements, but sometimes the width:100% was messing up other things (like the width of Wells). Additionally the first element in a form would have extra margin above it. And beyond that, the browser would give warnings for nested form elements. This PR replaces that need with FormElementContainer

## Types of changes

## Checklist:

### DIFF
--- a/src/view/components/formElementContainer.styl
+++ b/src/view/components/formElementContainer.styl
@@ -1,0 +1,7 @@
+.FormElementContainer > * {
+  margin-top: var(--spectrum-global-dimension-size-100);
+}
+
+.FormElementContainer > :first-child {
+  margin-top: 0;
+}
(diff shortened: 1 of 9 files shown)


jonsnyder:
@Aaronius, I also found a bug while testing the change here on the boolean XDMObject editor: https://jira.corp.adobe.com/browse/PDCL-6389

Aaronius (approved):
Thanks for doing this and logging the other issue you found.

src/view/components/formElementContainer.styl
```diff
@@ -0,0 +1,7 @@
+.FormElementContainer > * {
+  margin-top: var(--spectrum-global-dimension-size-100);
+}
+
+.FormElementContainer > :first-child {
+  margin-top: 0;
+}
```
Aaronius, line 7: I don't know why I didn't catch this when I reviewed the PR, but the proper way to do this would be to use the `gap` prop on the `Flex` component with a value of `size-100` (which in turn sets the `gap` CSS attribute). Feel free to change it right on the react-spectrum branch or I'll change it when I merge this stuff into my working changes.

PR #156, opened 2021-07-01
Fix broken test helper where off-screen pickers couldn't be clicked

jonsnyder:
## Description

Aaron was getting two test failures, but locally I was only getting one. Watching the failed test run, I could see that the picker wasn't being scrolled into view. I'm guessing that Aaron was getting another failure based upon the window size of his test runner. This change ensures the picker is scrolled into view before it is clicked on.

## Types of changes

## Checklist:


Aaronius (approved):
Nice find! They all pass on my machine.

PR #158, opened 2021-07-12
PDCL-6011 Migrate configuration view to react spectrum 3. Supports paging for edge configuration selection.

Aaronius:
## Types of changes

## Checklist:


.eslintrc.js
```diff
@@ -10,8 +10,10 @@ module.exports = {
     node: true,
     jasmine: true
   },
-  plugins: ["ban", "prettier", "testcafe"],
+  plugins: ["unused-imports", "ban", "prettier", "testcafe"],
   rules: {
+    "unused-imports/no-unused-imports": "error",
+    "unused-imports/no-unused-vars": "error",
```
Aaronius, line 16: This automatically removed unused imports. I got tired of doing it manually.

src/view/actions/setConsent.jsx
```diff
@@ -310,8 +309,7 @@ const ConsentObject = ({ value, index }) => {
           <RadioGroupWithDataElement
             dataTestIdPrefix="gdprContainsPersonalData"
             name={`consent[${index}].gdprContainsPersonalData`}
-            label="GDPR Contains Personal Data"
-            description="Does the event data associated with this user contain personal data?"
```
Aaronius, line 314: I changed FieldDescriptionAndError to derive its width based off its only child, rather than having to pass `width` to both `FieldDescriptionAndError` and the child of `FieldDescriptionAndError`.  I did this before you created RadioGroupWithDataElement. When I rebased on top of your changes, RadioGroupWithDataElement was feeding multiple children to FieldDescriptionAndError, so that was causing an error.

I debated on how to solve it, but I thought the UX was kind of cluttered with a description at the RadioGroupWithDataElement level:

image

So I removed FieldDescriptionAndError from RadioGroupWithDataElement and just made the labels be what the descriptions were.

src/view/errors/userReportableError.js
```diff
@@ -18,19 +18,14 @@ governing permissions and limitations under the License.
  */
 class UserReportableError extends Error {
   constructor(message, { originatingError, additionalInfoUrl } = {}) {
-    if (originatingError) {
-      // Log the originating error so that
-      // we can still see its stack trace.
-      console.error(originatingError);
-    }
-
```
Aaronius, line 26: See monitorForOriginatingErrors for why I removed this. TL;DR: if this UserReportableError is handled, it's confusing to have the originating error logged to the console.

src/view/utils/fetchFromEdge.js
```diff
@@ -0,0 +1,88 @@
...
+
+const ERROR_TYPE_INVALID_ACCESS_TOKEN =
+  "https://ns.adobe.com/aep/errors/EXEG-3032-401";
+
+export default async ({ orgId, imsAccess, path, params, headers, signal }) => {
+  const host = getHost({
+    imsAccess,
+    productionHost: EDGE_HOST_PROD,
+    stagingHost: EDGE_HOST_STAGING
+  });
```
Aaronius, line 30: Previously we were just using the prod endpoint for edge configurations, but I figured we should be consistent with what we're doing for the aep schema endpoints and support both staging and production.

src/view/dataElements/xdmObject/components/sandboxSelector.jsx
```diff
@@ -54,13 +53,11 @@ const SandboxSelector = ({
         }));
       } catch (e) {
         if (e.name !== "AbortError") {
-          reportAsyncError(
-            new UserReportableError(`Failed to load sandboxes.`, {
-              originatingError: e
-            })
-          );
-          return undefined;
+          reportAsyncError(e);
```
Aaronius, line 56: I moved the creating of UserReportableError up to fetchSandboxes (and did so similarly in other places). It seemed to make the most sense there and reduced duplicate code.

src/view/configuration/utils/getPartsFromEnvironmentCompositeId.js
```diff
@@ -10,22 +10,18 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  const [edgeConfigId, environmentName] = compositeId.split(":");
 
-  return abortController.signal;
+  return {
+    edgeConfigId,
+    environmentName
+  };
 };
 
-export default abortPreviousRequestsAndCreateNewSignal;
```
Aaronius, line 31: This diff is messed up; this wasn't a renamed file. `abortPreviousRequestsAndCreateNewSignal` was something I was using canceling requests in the XDM object data element. Notice that I had a "global" `abortController`, which was fine for the XDM object case, but it meant I couldn't re-use it for the  edge config selection in the configuration view. I decided to change it to a re-usable React hook, so the abortController would be scoped to the component using it. See `useAbortPreviousRequestsAndCreateSignal.jsx`.

test/functional/actions/sendEvent.spec.js
```diff
@@ -156,6 +153,7 @@ test("returns full valid settings with decision scopes as data element", async (
   });
   await instanceNameField.selectOption("alloy2");
   await typeField.enterSearch("mytype1");
+  await typeField.pressEnterKey();
```
Aaronius, line 156: I upgraded React-Spectrum's ComboBox and their new version requires hitting enter to accept the text instead of just blurring.

test/functional/actions/resetEventMergeId.spec.js
```diff
@@ -10,14 +10,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import createExtensionViewController from "../helpers/createExtensionViewController";
+import extensionViewController from "../helpers/extensionViewController";
 import spectrum from "../helpers/spectrum2";
 import createFixture from "../helpers/createFixture";
 
-const extensionViewController = createExtensionViewController(
-  "actions/resetEventMergeId.html"
```
Aaronius, line 18: The path here wasn't actually being used since I re-worked how tests are run. This means `createExtensionViewController` could just change to `extensionViewController`.

test/functional/configuration/configuration.spec.js
```diff
@@ -10,180 +10,76 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
   "\\.(exe|zip|wav|mp3|mov|mpg|avi|wmv|pdf|doc|docx|xls|xlsx|ppt|pptx)$";
 
-const defaultInitInfo = {
-  company: {
-    orgId: "ABC123@AdobeOrg"
-  },
-  tokens: {
-    imsAccess: "Badtoken"
-  }
-};
```
Aaronius, line 120: I moved this default init info into `extensionViewController`, which simplified some things.

test/functional/configuration/helpers/viewSelectors.js
```diff
@@ -0,0 +1,129 @@
...
+      webField: spectrum.checkbox("contextWebField"),
+      deviceField: spectrum.checkbox("contextDeviceField"),
+      environmentField: spectrum.checkbox("contextEnvironmentField"),
+      placeContextField: spectrum.checkbox("contextPlaceContextField")
+    },
+    deleteButton: spectrum.button("deleteInstanceButton")
+  });
+}
+
+export { instances };
```
Aaronius, line 129: I split these selectors into a separate file because (1) it was getting rather large and (2) I plan to re-use them in a spec file that tests edge configuration selection.

src/view/configuration/edgeConfigurationsSection.jsx
```diff
@@ -0,0 +1,397 @@
...
+        instanceValues.edgeConfigFreeformInputMethod = getFreeformInputStateForNewInstance();
+        instanceValues.edgeConfigInputMethod = INPUT_METHOD.SELECT;
+        isSuccessfullyPopulatedForSelectInputMethod = true;
+      } catch (e) {
+        // No nothing. We'll fall back to the freeform input method.
+      }
+    }
+
+    if (!isSuccessfullyPopulatedForSelectInputMethod) {
+      instanceValues.edgeConfigSelectInputMethod = await getSelectInputMethodStateForNewInstance(
```
jonsnyder, line 244: I don't understand this part. If this is the second instance there is no need to fetch configs. Indeed, it will return configs for the wrong IMS org.

src/view/configuration/edgeConfigurationsSection.jsx
```diff
@@ -0,0 +1,397 @@
...
+
+      if (!edgeConfigId) {
+        return true;
+      }
+
+      const firstInstanceWithSameEdgeConfigId = instances.find(
+        candidateInstance =>
+          getEdgeConfigIdFromInstance(candidateInstance) === edgeConfigId
+      );
+      const isDuplicate = firstInstanceWithSameEdgeConfigId !== instance;
```
jonsnyder, line 332: Clever!

src/view/configuration/edgeConfigurationsSection.jsx
```diff
@@ -0,0 +1,397 @@
...
+          >
+            Choose from list
+          </Radio>
+          <Radio
+            data-test-id="edgeConfigInputMethodFreeformRadio"
+            value={INPUT_METHOD.FREEFORM}
+          >
+            Enter values
+          </Radio>
+        </RadioGroup>
```
jonsnyder, line 378: Only the first instance should have this radio group. We cannot fetch the data streams for the second instance because the IMS org will be different.

jonsnyder (changes requested):
Lots of good stuff in here! My biggest concern is showing the data streams input method selector on the second instance.

src/view/actions/setConsent.jsx
Aaronius: 👍  Thanks. I'll update the docs.

Aaronius:
Thanks for the feedback! I've addressed your comments. Take a look at the last commit in particular. I created a method called `getMergedBridgeMethod` that's used for the various bridge methods. One side effect of this is that it makes `getSettings` asynchronous. Launch supports an asynchronous `getSettings`, and we could end up needing it eventually, so I decided to roll with it. It does change spectrum3ExtensionView.jsx, which might conflict with what you're working on.

jonsnyder:
I found some issues while testing the UI:

1. The descriptions on the privacy options are missing
2. I think the onBeforeEventSend label should be "On before event send callback". For consistency we have not used the property name as the label before.
3. Required field asterisks are not consistent. (IMS Org Id, Name, Edge Domain, Default Consent, Edge base path, Consistency between datastreams select and form, download link qualifier)
4. "Restore to default" and "Test Regex" buttons are not clearly buttons. In the old UI, the hover state for these showed the outline of a button.
5. Too much space between Enable Click Collection checkbox and Download Link Qualifier
6. In extension.json, downloadLinkQualifier has a minimum length of 1, but there is no such validation in the UI

src/view/configuration/configuration.jsx
```diff
@@ -79,33 +97,11 @@ const getInitialValues = async ({ initInfo }) => {
   if (instancesSettings) {
     instancesInitialValues = await Promise.all(
       instancesSettings.map(async (instanceSettings, instanceSettingsIndex) => {
```
jonsnyder, line 99: No longer need this async.

jonsnyder (approved):
Looks good. Thanks!

src/view/components/formikReactSpectrum3/radioGroupWithDataElement.jsx
```diff
@@ -38,7 +38,6 @@ const RadioGroupWithDataElement = ({
   name,
   children,
   dataElementDescription,
-  width,
```
Aaronius, line 41: Having `width` on this component didn't really do what was expected because we were forwarding on the width to the ReactSpectrumRadioGroup and the data element textfield, which is indented. Meaning the total width of the component ends up being larger than defined width. Everywhere we're currently using RadioGroupWithDataElement seems to look better without a constrained width as well. Because of these things, I just decided to remove the prop entirely and if we want to add support for width later, we'll need to be smarter about how we accomplish it.

src/view/components/formikReactSpectrum3/radioGroupWithDataElement.jsx
```diff
@@ -141,15 +140,16 @@ const RadioGroupWithDataElement = ({
...
                 value={dataElementText}
                 data-test-id={`${dataTestIdPrefix}DataElementField`}
                 onChange={newValue => setValues("dataElement", newValue)}
                 onBlur={() => dataElementSetTouched(true)}
                 validationState={
                   dataElementTouched && error ? "invalid" : undefined
                 }
-                width={width}
+                width="size-5000"
+                isRequired
```
Aaronius, line 152: When the "data element" radio is selected, this textfield becomes required, so I think it makes sense to mark the textfield as required. The required indicator only shows up if we have a label though, which is why I switched the prop from `aria-label` to `label`. I added labels to other textfields that didn't have labels to be consistent.

src/view/configuration/dataCollectionSection.jsx
```diff
@@ -146,60 +146,62 @@ const DataCollectionSection = ({ instanceFieldName }) => {
...
+                </ActionButton>
+                <RestoreDefaultValueButton
+                  data-test-id="downloadLinkQualifierRestoreButton"
+                  name={`${instanceFieldName}.downloadLinkQualifier`}
+                  defaultValue={instanceDefaults.downloadLinkQualifier}
+                />
+              </Flex>
+            </FieldSubset>
+          )}
+        </div>
```
Aaronius, line 204: Wrapped in a div because FormElementContainer was adding space between the checkbox and regex field, causing the additional spacing you mentioned.

src/view/dataElements/xdmObject/components/objectEdit.jsx
```diff
@@ -55,7 +55,8 @@ const ObjectEdit = ({ fieldName }) => {
             <TextField
               data-test-id="valueField"
               name={`${fieldName}.value`}
-              label="Data element providing array"
```
Aaronius, line 58: Oops. This should have said "object".

Aaronius:
Thanks for the finds, @jonsnyder! I've made the updates. Regarding marking fields as required, I talked to the Spectrum folk and they said:

> No need to mark it as required if the user can't not fill it out.

I decided to go with that approach.

Aaronius:
Thanks for the finds, @jonsnyder! I've made the updates. Regarding marking fields as required, I talked to the Spectrum folk and they said:

> No need to mark it as required if the user can't not fill it out.

I decided to go with that approach.

Aaronius:
@jonsnyder I'm going to merge this, but if you review the latest changes and have feedback, I'm happy to address them in a later PR. I have an upcoming PR for removing React-Spectrum 2 stuff, so making additional changes in this branch is making it more difficult to keep that other branch properly rebased.

src/view/actions/setConsent.jsx
Aaronius: @jonsnyder When you mentioned that this change requires a change to the documentation, I believe you were thinking that the labels were referenced in the documentation and those references needed be changed. Just so you're aware, I didn't see any label references from the documentation (I assume they would be [here](https://experienceleague.adobe.com/docs/experience-platform/edge/extension/action-types.html?lang=en#set-consent), so I don't think anything needs to be changed. If you're aware of where they may be referenced or if you were thinking about something different when you made your comment, let me know.

PR #159, opened 2021-07-15
React-Spectrum 2 Cleanup

Aaronius:
## Description
Now that we've implemented React-Spectrum 3, this cleans up all the React-Spectrum 2 related files. Also, since React-Spectrum 3 provides props for handling most layout styling, I went through the process of removing our `utils.styl` and leveraging props instead. We may still decide to re-add a `utils.styl` in the future for things that props don't cover. The one util I see as potentially still being useful is `u-noWrap` because that isn't yet supported through React-Spectrum 3 props, but [I'm hoping that changes](https://github.com/adobe/react-spectrum/issues/2105).

I hope everything else is clear. Let me know if you have questions.

## Types of changes

## Checklist:


package.json
```diff
@@ -45,20 +45,17 @@
     "@adobe/react-spectrum": "^3.11.0",
     "@react-spectrum/combobox": "^3.0.0-rc.0",
     "@react-stately/data": "^3.4.0",
-    "@react/react-spectrum": "^2.34.0",
```
jonsnyder, line 48: 🕺

src/view/components/formikReactSpectrum3/radioGroupWithDataElement.jsx
```diff
@@ -12,11 +12,7 @@ governing permissions and limitations under the License.
```
jonsnyder, line 12: Following the naming conventions so far this should probably be named "src/view/components/formikReactSpectrum3/formikRadioGroupWithDataElement.jsx"

jonsnyder (approved):

src/view/components/formikReactSpectrum3/radioGroupWithDataElement.jsx
Aaronius: Ah, yes. This wasn't intentional. Thanks for bringing it up.

PR #160, opened 2021-07-16
PDCL-5969 - add a new event Send Event Complete

ninaceban:
## Description

Added a new Event - Send Event Complete
It will have in the payload the whole object that the `sendEvent` command is resolved with.
Added `DEPRECATED` label to the old Decisions Received Event.
## Types of changes

## Checklist:

### DIFF
--- a/src/lib/sendEventCallbackStorage.js
+++ b/src/lib/sendEventCallbackStorage.js
@@ -16,9 +16,9 @@ module.exports = {
   add(trigger) {
     sendEventCompleteTriggers.push(trigger);
   },
-  triggerEvent(resultingObject) {
+  triggerEvent(result) {
     sendEventCompleteTriggers.forEach(trigger => {
-      trigger(resultingObject);
+      trigger(result);
     });
   }
 };
--- a/src/lib/sendEventCallbackStorage.js
+++ b/src/lib/sendEventCallbackStorage.js
@@ -1,0 +1,24 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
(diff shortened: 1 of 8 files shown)


Aaronius:
Hey @ninaceban, it's looking good, but I think you left out the part where the Send Event action calls `sendEventCallbackStorage.triggerEvent` with the `result` object.

src/lib/sendEventCallbackStorage.js
```diff
@@ -0,0 +1,24 @@
...
+governing permissions and limitations under the License.
+*/
+
+const sendEventCompleteTriggers = [];
+
+module.exports = {
+  add(trigger) {
+    sendEventCompleteTriggers.push(trigger);
+  },
+  triggerEvent(resultingObject) {
```
Aaronius, line 19: I think just `result` instead of `resultingObject` would be sufficient.

test/unit/lib/actions/sendEvent/createSendEvent.spec.js
```diff
@@ -72,6 +77,9 @@ describe("Send Event", () => {
       expect(decisionsCallbackStorage.triggerEvent).toHaveBeenCalledWith({
         decisions
       });
+      expect(sendEventCallbackStorage.triggerEvent).toHaveBeenCalledWith({
+        decisions
+      });
```
Aaronius, line 82: Let's improve this assertion. In the real world, `triggerEvent` will now be called with an object containing both `decisions` and `propositions`, and we want to make sure that `sendEventCallbackStorage.triggerEvent` has been called with an object containing both `decisions` and `propositions` and that `decisionsCallbackStorage.triggerEvent` has been called with an object just containing `decisions`. You'll need to update our mock instance [here](https://github.com/adobe/reactor-extension-alloy/pull/160/commits/fcf907508ae696de24dfef217304876fcf2d7721#diff-17378a7b78223802022d653d7ae99071e331af5f9520eb536e32ad6fda9eb2afR18-R20) so that it resolves with a promise that contains both `decisions` and `propositions`, like Alloy now behaves.

Aaronius (approved):
Looks good, @ninaceban. I think we should improve the one assertion I pointed out. Otherwise, I think it's ready to 🚢 .

jfkhoury (approved):
Nice work!

test/unit/lib/actions/sendEvent/createSendEvent.spec.js
ninaceban: good point! thank you

PR #161, opened 2021-07-16
Made XDM object edit form sticky.

Aaronius:
## Description
If you expand the schema tree within the XDM Object data element view so that it becomes really long, then you scroll down, the edit form on the right side is scrolled up out of sight. This makes for an awkward experience, especially when you select a node on the XDM tree and expect something to change on the right side of the view. This really bugged me, so I decided to fix. Now the edit form on the right side is "sticky" and won't scroll out of view.

I also fixed a spacing inconsistency between the array edit form and the object edit form.

This is with the sticky edit form:
image

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/xdmObject/components/arrayEdit.jsx
+++ b/src/view/dataElements/xdmObject/components/arrayEdit.jsx
@@ -23,6 +23,7 @@ import getInitialFormState, {
 } from "../helpers/getInitialFormState";
 import { PARTS, WHOLE } from "../constants/populationStrategy";
 import { ARRAY, OBJECT } from "../constants/schemaType";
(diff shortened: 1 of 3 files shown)


jfkhoury (approved):
Love this change!

jonsnyder (commented):
Great idea! I found a couple bugs. (I was running on Safari)

1. When the tree is taller than the window, In order to scroll the tree, you have to have your pointer in the right pane.
2. Scrollbars appear on the tree before it is taller than the window. (See screenshoot)
Screen Shot 2021-07-20 at 9 28 42 AM

Aaronius:
@jonsnyder and I discussed this more and I added a commit to work around scroll issues (and provide a potentially better UX).

PR #162, opened 2021-07-16
Remove multiple formiks

jonsnyder:
## Description

Simplify the extension views by removing the multiple Formik states.

## Motivation and Context

As part of the react-spectrum changes I added support for having multiple Formik states rendered at once. This made the extension views more complicated though.

## Types of changes

## Checklist:


src/view/actions/resetEventMergeId.jsx
```diff
@@ -40,60 +37,28 @@ const validationSchema = object().shape({
...
-          <ExtensionViewForm
-            render={props => {
-              return <ResetEventMergeId {...props} />;
-            }}
+      getInitialValues={getInitialValues}
+      getSettings={getSettings}
+      formikStateValidationSchema={validationSchema}
+      render={() => (
+        <FormElementContainer>
+        <DataElementSelector>
```
Aaronius, line 48: Looks like indentation needs adjusted. If you run `npm run format`, you'll see one other adjustment in a different file as well. Running `npm run lint` will remove a couple unused imports as well.

src/view/components/decisionScopes.jsx
```diff
@@ -183,21 +162,3 @@ const DecisionScopes = ({
     </div>
   );
 };
-
-DecisionScopes.propTypes = {
-  initInfo: PropTypes.object,
-  formikProps: PropTypes.object,
-  registerImperativeFormApi: PropTypes.func
-};
```
Aaronius, line 191: This is interesting. Apparently DecisionScopes is no longer recognized as a React component, maybe because it's an anonymous function? I think I'd prefer keeping it as a named function for easier debugging when there's an error in the component, but what do you think? Also, do you feel like prop types are providing value for the cost of maintenance?

src/view/components/extensionView.jsx
```diff
@@ -10,76 +10,170 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-// ExtensionViewContext. It should be used for each view inside an extension.
-const ExtensionView = ({ render }) => {
+const ExtensionView = ({
+  render,
+  getInitialValues,
+  getSettings,
+  validateFormikState,
+  formikStateValidationSchema,
+  validateNonFormikState
+}) => {
```
Aaronius, line 30: I like that you support both props and `registerImperativeFormApi`. That's kind of nice.

src/view/configuration/configuration.jsx
```diff
@@ -138,32 +134,10 @@ const validationSchema = object().shape({
...
-    });
-    let initialValues;
-    try {
-      initialValues = await getInitialValues({
-        initInfo
-      });
-    } catch (e) {
-      reportAsyncError(e);
-      return;
-    }
```
Aaronius, line 163: I think we need to take into consideration error handling for the `getInitialValues` call that we're now doing in ExtensionView. If anything goes wrong in `getInitialValues`, we should report an async error, I believe.

src/view/dataElements/identityMap.jsx
```diff
@@ -407,30 +383,19 @@ function IdentityMap({ initInfo, formikProps, registerImperativeFormApi }) {
...
-    </React.Fragment>
+    </>
   );
-}
-
-IdentityMap.propTypes = {
-  initInfo: PropTypes.object,
-  formikProps: PropTypes.object,
-  registerImperativeFormApi: PropTypes.func
 };
```
Aaronius, line 388: How did you get away with removing prop types?

Aaronius (approved):
Good changes!

src/view/components/decisionScopes.jsx
jonsnyder: I think prop types are valuable. It has led to better error messages and easier debugging for me on a number of occasions.

src/view/dataElements/identityMap.jsx
jonsnyder: I removed all the props on the component so there is no need to declare propTypes.

src/view/actions/sendEvent.jsx
```diff
@@ -14,13 +14,11 @@ import React from "react";
 import { object, string } from "yup";
 import { Item } from "@adobe/react-spectrum";
 import FormikComboBox from "../components/formikReactSpectrum3/formikComboBox";
-import FormikPicker from "../components/formikReactSpectrum3/formikPicker";
 import FormikTextField from "../components/formikReactSpectrum3/formikTextField";
 import FormikCheckbox from "../components/formikReactSpectrum3/formikCheckbox";
 import DataElementSelector from "../components/dataElementSelector";
 import render from "../render";
 import ExtensionView from "../components/extensionView";
-import getInstanceOptions from "../utils/getInstanceOptions";
```
Aaronius, line 24: Some of these were definitely my fault.

src/view/components/extensionView.jsx
```diff
@@ -93,15 +93,19 @@ const ExtensionView = ({
     init({ initInfo: _initInfo }) {
       setInitInfo(_initInfo);
     },
-    getSettings() {
+    async getSettings() {
```
Aaronius, line 96: Good catch.

Aaronius:
Yeah, that all makes sense. Looks good!

src/view/components/extensionView.jsx
```diff
@@ -93,15 +93,19 @@ const ExtensionView = ({
...
-        values: formikPropsRef.current.values
-      });
+      try {
+        return await viewRegistrationRef.current.getSettings({
+          initInfo,
+          values: formikPropsRef.current.values
+        });
+      } catch (e) {
+        reportAsyncError(e);
+        return {};
```
Aaronius, line 107: I guess I'm not entirely sure what the return value should be here. If we return an empty object, I believe Launch will try to validate it against the schema in our extension.json. If we have any required fields, that would prevent the empty object from actually being saved. But, if we don't have any required fields, then I think an empty object would be saved, which wouldn't be good because it wouldn't represent the actual form values. Re-throwing the error might actually be best.

PR #166, opened 2021-07-21
Adding production build support for nullish coalescing operator and optional chaining.

Aaronius:
## Description
When generating a production build, Terser would fail. Terser is responsible for minification. The Terser that's bundled with Parcel is an older version, which didn't support the nullish coalescing operator or optional chaining. Unfortunately, Parcel hasn't updated their Terser dependency. Parcel 2, which is yet to be released, will use the new Terser and this problem will go away.

The way I chose to get around this is by forcing Babel to transpile nullish coalescing operators and optional chaining. This way, the code that gets fed into Terser for minification doesn't have these usages and we won't get any error. It also means that the emitted build will have nullish coalescing operators and optional chaining transpiled, when they don't really need to be transpiled to operate properly in Launch's target browsers, but I don't think it's a big deal.

I wanted to leave a comment in `.babelrc` explaining this, but it's a JSON file, so that's hard to do. I tried converting it from a JSON file to a `babel.config.js` so that I could add a comment, but Parcel doesn't support a `babel.config.js` either (though Parcel 2 does). 😢 

## Types of changes

## Checklist:

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,11 @@
   "presets": [["@babel/preset-env"]],
   "env": {
     "production": {
-      "plugins": ["babel-plugin-jsx-remove-data-test-id"]
+      "plugins": [
+        "babel-plugin-jsx-remove-data-test-id",
+        "@babel/proposal-nullish-coalescing-operator",
+        "@babel/proposal-optional-chaining"
+      ]
     }
   }
 }


jonsnyder (approved):
I was looking at upgrading to Parcel 2. It is in beta, but has been for some time. I think this is a much better solution. We can wait for Parcel 2 to be finalized.

Aaronius:
I actually started upgrading to Parcel 2, but I found the [documentation for the library API](https://v2.parceljs.org/features/parcel-api/) inadequate. I then tried to just see if I could get the CLI working and I ran into another error for which there's an open issue, so I bailed.

PR #169, opened 2021-07-22
Automate releases

jonsnyder:
## Description

Implements the project card release automation for the extension.

## Types of changes

## Checklist:

### DIFF
--- a/scripts/updateAlloy.js
+++ b/scripts/updateAlloy.js
@@ -1,0 +1,37 @@
+#!/usr/bin/env node
+
+// Updates alloy version in package.json. For example:
+// 2.5.0-alpha.0 => 2.5.0-beta.4
+// 2.5.0-beta.4 => 2.5.0
+
+// This will not update to the next major/minor/patch version,
+// but only update the prerelease version or remove the prerelease
+// qualifier. So it will NOT do this:
+// 2.5.0-beta.4 => 2.6.0-beta.0
+// 2.5.0 => 2.6.0-beta.0
+
+// We do this instead of npm update so that we are explicit about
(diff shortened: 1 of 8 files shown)


scripts/updateAlloy.js
```diff
@@ -0,0 +1,37 @@
...
+
+const {
+  dependencies: {
+    "@adobe/alloy": { version: currentVersion }
+  }
+} = JSON.parse(execSync("npm ls @adobe/alloy --json").toString());
+
+// fetch any releases greater than or equal to the current version, with the current major/minor/patch number.
+const npmView = JSON.parse(
+  execSync(`npm view @adobe/alloy@~${currentVersion} version --json`).toString()
```
Aaronius, line 26: Is this assuming that the Alloy we want to ship is always within the same minor version of the extension (I thought that's what `~` means)? If so, I'm not sure that will be the case. From what we've discussed previously, Alloy's version and the extension version could be different minor versions (and potentially different major versions even?). For example, if we add a feature to the extension but don't change Alloy, the extension minor version would be bumped, but not Alloy's.

Maybe we should just prompt for an Alloy version when the developer starts the workflow?

jonsnyder: "currentVersion" is the current version of **Alloy** installed in package.json, not the version of the extension. I've started writing documentation for how this would work here: https://github.com/adobe/reactor-extension-alloy/wiki/Publishing-a-New-Version

Aaronius (approved):
I think it makes sense. I'll get a better handle on it as we use it in our process.

jfkhoury (approved):
Can’t wait to see the demo :)

PR #172, opened 2021-07-30
Improve typography.

Aaronius:
## Description
While I was working on the XDM merging feature, I learned a couple things.

1. React-Spectrum doesn't load Adobe Clean, so the font being displayed wasn't actually Adobe Clean.
2. React-Spectrum's Heading component doesn't actually provide heading styles. The Heading component is really for indicating which piece of content is intended to be the heading within a component like Dialog or IllustratedMessage. The Heading receives its styling from the parent component. It's not intended to be used on its own.

I talked to the React-Spectrum team about this and they said that typography components are still in the works and suggested that we make our own components as needed based on Spectrum CSS. Spectrum CSS typography is outlined [here](https://opensource.adobe.com/spectrum-css/typography.html).

We're now loading Adobe Clean. I've also created Heading and Body components to be used where needed. The Body component really only needs to be used where we want a font size that's different than the default (React-Spectrum applies default body styles to everything within the React-Spectrum Provider, which entails all of our content).

## Motivation and Context
Proper typography according to Spectrum.

## Types of changes

## Checklist:


src/view/components/alert.styl
```diff
@@ -20,10 +20,6 @@
   border-color: var(--spectrum-semantic-negative-color-default);
 }
 
-.Alert-title {
-  margin: 0
-}
```
Aaronius, line 25: The styles Spectrum CSS provides for our Heading component specify no margin, which allowed me to remove this.

src/view/configuration/configuration.jsx
```diff
@@ -20,7 +20,7 @@ import {
   DialogTrigger,
   Dialog,
   Flex,
-  Heading,
+  Heading as HeadingSlot,
```
Aaronius, line 23: Note that there is now React-Spectrum's `Heading` component and our own typography `Heading` component. I considered naming our typography `Heading` component something like `TypographyHeading` to distinguish. I'm still open to something like that.

src/view/components/typography/body.jsx
```diff
@@ -0,0 +1,59 @@
...
+ * This is typically unnecessary, since the React-Spectrum Provider
+ * component provides basic default body styling for all content,
+ * but this can be useful in scenarios where a different font size
+ * is needed.
+ * See https://opensource.adobe.com/spectrum-css/typography-body.html
+ * for more info. Once React-Spectrum provides a Body component
+ * that handles typography, we can get rid of this.
+ */
+const Body = ({ children, size = "S", isSerif, marginTop, marginBottom }) => {
+  const style = {};
```
jfkhoury, line 29: Should we allow passing the whole style? too much?

jfkhoury (approved):
Nice!!! Does this require any tests?

src/view/components/typography/body.jsx
Aaronius: I think it would be best to wait until we need it. If we do end up needing it, I believe we should expose it as a prop named `UNSAFE_style`, because that's how React-Spectrum would do it, and I hope to replace this eventually with an actual React-Spectrum implementation. Typically, React-Spectrum provides enough props related to styling that you don't need to use the `UNSAFE_style` prop. You can see a list of all their typical styling props [here](https://react-spectrum.adobe.com/react-spectrum/View.html#props). To keep things simple in our implementation, I've only exposed `marginTop` and `marginBottom` from that list, because I think that's all we would typically need in this component.

Aaronius:
> Nice!!! Does this require any tests?

It's a good question. Our functional tests are currently testing extension views rather than individual components. I've thought about deviating from this (loading up an HTML page that just renders a single component), but our extension view tests typically cover enough that creating tests for individual components hasn't seemed necessary. If we wanted to stick with testing extension views rather than individual components, I could add some tests to a spec that's testing a view that's using these components. Anyway, I don't feel like testing strictly presentational items provides much ROI. For example, we can test that an `h2` was produced and has a particular class name applied, but does it actually look reasonable?

I'm on the fence. I'm curious to hear what @jonsnyder thinks about this.

shammowla (approved):

jonsnyder:
I would love to be able to test individual components in the functional tests. It would require a different testing harness though. Currently we run everything through the extension sandbox which doesn't allow us to load up individual components. I don't think this PR warrants going though all that. This PR should probably add a test that the correct font is being used on one of the views.

The ROI I see in testing components in isolation simplifies the tests so that testing requires less code. For example, I see this being particularly useful in testing the paged combo boxes on the XDM view. Currently we create a bunch of web mocks to do this testing, but if we could test it in isolation we could provide promises instead of having to create web mocks. Then in the XDM view tests we can be fairly confident that the paging will work correctly. Component tests would also be helpful in the formik wrapped components, especially the radio group with data element.

jonsnyder (approved):

Aaronius:
Do you mind taking another look? I added a test for determining whether Adobe Clean was loaded. I decided to run it for all views.

Regarding the desire to test individual components, we're no longer running functional tests using the extension sandbox (I removed that as part of the react-spectrum upgrade work), so it shouldn't require much. I'm going to open a separate pull request to show what that would look like.

jonsnyder (approved):

PR #173, opened 2021-08-02
Removed decisions callback storage.

Aaronius:
## Description

We had both a `sendEventCallbackStorage` and a `decisionsCallbackStorage`. After receiving a response from the server, we would always notify the `sendEventCallbackStorage` but only notify the `decisionsCallbackStorage` if we actually received decisions from the server.

Having two callback storages made things more complicated than they needed to be. I've removed the `decisionsCallbackStorage`. The Decisions Received event type just listens to the `sendEventCallbackStorage` now. When it hears that an event has completed, it looks at the result to see if it finds decisions. If decisions are found, then it triggers the rule; otherwise, it does not trigger the rule.

Note that previously we were always accessing the `result.decisions` property after every sendEvent command. With this PR, we'll only be accessing `result.decisions` if the customer is actually using the Decisions Received event type. If the customer isn't using the Decisions Received event type anywhere in their Launch property, then [the Decisions Received event type JavaScript module won't even be included in the Launch library](https://medium.com/adobetech/build-optimization-through-minimal-inclusion-87ea637e174f). This plays into, but doesn't entirely solve, https://jira.corp.adobe.com/browse/PDCL-6567.  If we choose to fix PDCL-6567 by removing the `decisions` deprecation warning entirely, then this pull request isn't necessary, but still desirable.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-6567

## Motivation and Context
Simplify code, partially solve https://jira.corp.adobe.com/browse/PDCL-6567

## Types of changes

## Checklist:


src/lib/dataElements/eventMergeId/index.js
```diff
@@ -16,6 +16,5 @@ const eventMergeIdCache = require("../../eventMergeIdCache");
 
 module.exports = createEventMergeId({
   instanceManager,
-  eventMergeIdCache,
-  turbine
```
Aaronius, line 20: Unrelated, but `turbine` wasn't being used.

test/unit/lib/events/decisionsReceived/createDecisionsReceivedEvent.spec.js
```diff
@@ -1,31 +0,0 @@
...
-
-    const decisionsReceived = createDecisionsReceivedEvent({
-      decisionsCallbackStorage
-    });
-
-    decisionsReceived({}, trigger);
-
-    expect(decisionsCallbackStorage.add).toHaveBeenCalledWith(trigger);
-  });
-});
```
Aaronius, line 31: This wasn't really deleted. It was just renamed and modified.

src/lib/events/decisionsReceived/createDecisionsReceived.js
```diff
@@ -10,6 +10,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
-module.exports = ({ decisionsCallbackStorage }) => (settings, trigger) => {
-  decisionsCallbackStorage.add(trigger);
+module.exports = ({ sendEventCallbackStorage }) => (settings, trigger) => {
+  sendEventCallbackStorage.add(result => {
+    if (result.decisions) {
+      trigger({ decisions: result.decisions });
+    }
+  });
 };
```
Aaronius, line 19: I renamed this file to be consistent with what we've done elsewhere. That is, we don't suffix the files with `Event`, `Action`, `DataElement`, etc.

jonsnyder (approved):
It's always good to see less code with the same functionality.

jfkhoury (approved):
LOVE Those red lines!! Thank you!

PR #178, opened 2021-08-11
Added typography component tests and foundation for future component tests.

Aaronius:
## Description
Added typography component tests and foundation for future component tests.

## Motivation and Context
Improved testing for more robust software.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/components/typography/body.styl
+++ b/src/view/components/typography/body.styl
@@ -1,0 +1,15 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+// Body classes are not bundled with React-Spectrum, so we have
+// to import them directly.
+@import "@spectrum-css/typography/dist/index-vars.css";
--- a/test/functional/actions/resetEventMergeId.spec.js
+++ b/test/functional/actions/resetEventMergeId.spec.js
@@ -12,12 +12,12 @@ governing permissions and limitations under the License.
 
 import extensionViewController from "../helpers/extensionViewController";
 import spectrum from "../helpers/spectrum3";
-import createFixture from "../helpers/createFixture";
+import createExtensionViewFixture from "../helpers/createExtensionViewFixture";
 import runCommonExtensionViewTests from "../runCommonExtensionViewTests";
 
 const eventMergeIdField = spectrum.textField("eventMergeIdField");
 
-createFixture({
+createExtensionViewFixture({
   title: "Reset Event Merge ID View",
   viewPath: "actions/resetEventMergeId.html"
 });
--- a/test/functional/components/typography/body.fixture.html
+++ b/test/functional/components/typography/body.fixture.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Component View</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script src="body.fixture.jsx"></script>
+  </body>
+</html>
--- a/test/functional/components/typography/body.spec.js
+++ b/test/functional/components/typography/body.spec.js
@@ -1,0 +1,89 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { t } from "testcafe";
+import createComponentFixture from "../helpers/createComponentFixture";
+import { createTestIdSelector } from "../../helpers/dataTestIdSelectors";
+
+createComponentFixture({
+  title: "Body component",
+  htmlFileName: "body.fixture.html"
+});
+
+const scenarios = [
+  {
+    testId: "sizeXxxl",
+    classNames: ["spectrum-Body--sizeXXXL"]
+  },
+  {
+    testId: "sizeXxl",
+    classNames: ["spectrum-Body--sizeXXL"]
+  },
+  {
+    testId: "sizeXl",
+    classNames: ["spectrum-Body--sizeXL"]
+  },
+  {
+    testId: "sizeL",
+    classNames: ["spectrum-Body--sizeL"]
+  },
+  {
+    testId: "sizeM",
+    classNames: ["spectrum-Body--sizeM"]
+  },
+  {
+    testId: "sizeS",
+    classNames: ["spectrum-Body--sizeS"]
+  },
+  {
+    testId: "sizeXs",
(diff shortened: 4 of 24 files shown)


src/view/components/typography/body.styl
```diff
@@ -0,0 +1,15 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+// Body classes are not bundled with React-Spectrum, so we have
+// to import them directly.
+@import "@spectrum-css/typography/dist/index-vars.css";
```
Aaronius, line 15: Technically, not having this file was a bug. If the Body component were used, the Body spectrum-css files wouldn't be included and therefore the `spectrum-Body` class (and other associated Body classes) wouldn't be included or applied. However, we're only using Body in one place at the moment, and it's used on a view that also uses Heading, and Heading also imports this same `index-vars.css` file, which is why we weren't/aren't seeing issues.

This actually shows a bit of a weakness in the Heading and Body tests that are included in this PR, because none of these new tests are testing that the computed styles the element is receiving are coming from the `index-vars.css` file; we're just testing that we are setting the right classes and styles on the element.

One thing we _could_ do is pick a specific style that should _only_ be coming from `index-vars.css` and test that the element ends up with a computed style that matches. At that point, though, we're kind of testing internals of `index-vars.css` (if style values inside index-vars.css change, our test may break) and also assuming that the computed style will continue being uniquely sourced from `index-vars.css` versus something else in React-Spectrum.

I'm okay with just testing that we're setting the right class and styles on the element as I've done in this PR and dealing with the risk trade-off, but I'm also open to other opinions.

test/functional/actions/resetEventMergeId.spec.js
```diff
@@ -12,12 +12,12 @@ governing permissions and limitations under the License.
...
 import extensionViewController from "../helpers/extensionViewController";
 import spectrum from "../helpers/spectrum3";
-import createFixture from "../helpers/createFixture";
+import createExtensionViewFixture from "../helpers/createExtensionViewFixture";
 import runCommonExtensionViewTests from "../runCommonExtensionViewTests";
 
 const eventMergeIdField = spectrum.textField("eventMergeIdField");
 
-createFixture({
+createExtensionViewFixture({
```
Aaronius, line 20: I renamed this to distinguish it from `createComponentFixture`.

test/functional/components/typography/body.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      .eql(["spectrum-Body"].concat(classNames));
+    if (style) {
+      await t.expect(selector.getAttribute("style")).eql(style);
+    } else {
+      await t
+        .expect(selector.getAttribute("style"))
+        .notOk("Style found when no style was expected.");
+    }
+  });
+});
```
Aaronius, line 89: Notice that our `forEach` runs a `test` on each iteration. TestCafe reloads the page for each `test`, making it take longer to complete than necessary. There are ways to get TestCafe to not reload the page between tests, particularly by setting `disablePageReloads` on the fixture, but with it comes a downside: if you're running `npm run test:functional:watch` and change a fixture files (`body.fixture.jsx` or `body.fixture.html`) TestCafe won't ever pick up the latest fixture file changes, because TestCafe never reloads the page, even when you're explicitly telling it to restart the test run by hitting Ctrl+R. 😢  It would be great if TestCafe re-loaded the page between each test run, but it doesn't. 🤷  

Alternatively, I could move the `forEach` to _inside_ the `test` so the page is only loaded once per test run, but then if one of the scenarios fails, it's more difficult to figure out which one failed. There are ways to mitigate this but logging the scenario for each assertion. Something like:

```
await t
  .expect(selector.getAttribute("style"))
  .notOk("Style found when no style was expected for scenario `${testId}`.");
```

But it's kind of gross to add that to each assertion. As another alternative, we could log the scenario on each iteration using `console.log`, which might help.

Annnyway....I just decided to stick with the slower, more orthodox approach.

src/view/components/typography/body.styl
jonsnyder: Having a Body component isolated test would find such problems, but that may be out of scope for this PR.

Aaronius: I think we may be miscommunicating and it's probably my fault. This PR does have a test that isolates the Body component, if I understand what you mean. The test, though, asserts that we apply the right classes to the element, not that the computed font size/style/color is a specific value being provided by `index-vars.css`. In other words, the test would still pass even if `index-vars.css` weren't loaded by Body.

jfkhoury (approved):
Cool change!

test/functional/components/typography/body.fixture.html
```diff
@@ -0,0 +1,11 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Component View</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script src="body.fixture.jsx"></script>
+  </body>
+</html>
```
jonsnyder, line 11: Is it possible to include jsx in the test file? It seems like a lot of work to create a .fixture.html, .fixture.jsx, and .spec.js file for every component we want to test. For example, you could save duplication by producing a body component instance within the test as you already have an array of test cases with all the sizes. Do you see the .fixture.html file always being the same except loading a different jsx file?

jonsnyder (approved):
I'm excited to start building component tests!

test/functional/components/typography/body.fixture.html
Aaronius: I considered this in my approach and the answer is yes, you can use JSX inside TestCafe specs, but with limitations. For example, if the JSX in the spec file has `<MyCustomComponent>`, then I need to import `MyCustomComponent` into the spec file and rely on TestCafe's transpilation of `MyCustomComponent`. However, TestCafe's transpilation doesn't use the babel configuration that we've set up for our project, so it ends up choking on some files, like our Stylus files. This predicament is explained pretty well by [this comment](https://github.com/DevExpress/testcafe/issues/2659#issuecomment-492436113) (if that link doesn't take you to the comment, look for the comment by anonmos).

Even if TestCafe were to be able to transpile our JSX properly, that would tranpile the JSX within the _spec_ file, which runs in Node. We need it to transpile it in such a way that it can actually render react elements on the page.  Putting the JSX in a `ClientFunction` would work if we loaded React and ReactDOM onto the fixture HTML page, but we still wouldn't be able to use our custom components within the JSX because we'd still be importing the custom components within our spec file, when we really need the custom component code on the fixture page itself (and we run into the transpilation problem I just described anyway). We'd only be able to render regular DOM elements, because those are provided by ReactDOM, which would be on the fixture page. Hopefully that makes sense.

Your questioning made me think through this problem more though. I've added a commit that solves the issue in a different way. It doesn't allow us to use JSX with custom components in our spec file, but I don't think that's the important part anyway. I think the important part is to have the rendering logic for each test case inside the spec file instead of the fixture.

In the commit I added, we use a single fixture HTML file. The fixture HTML file "imports" a single JSX file which imports all our custom components we want to test. It also adds a global method on window (`window.renderSerializedReactElement`). In our spec file, _outside_ of any ClientFunction, we can call `React.createElement("MyCustomComponent", props, children`)`. Now, when using `React.createElement`, you typically pass the actual component function as the first argument. In our case, we're passing a string, which is important. When the test runs, React.createElement will run, which spits out a plain JavaScript object representing the element graph that needs to be rendered, but it retains the string `MyCustomComponent` for the type (remember, we can't import the `MyCustomComponent` component inside the spec file). In our case, we need to render the element graph on the _page_, not in the _spec file_, so we send that element graph object to the page using a `ClientFunction` we've defined. The `ClientFunction` then, on the fixture page, passes the element graph to `window.renderSerializedReactElement` (which our fixture has provided). `window.renderSerializedReactElement` then "deserializes" the object and swaps out the `MyCustomComponent` string (which is the value for the `type` property on the element graph object) with the actual `MyCustomComponent` component function. We then feed that deserialized element graph to `React.render`, which then renders the content to the page.

The downside about this approach is that when we want to test a new component, we need to import that component in our fixture file. I added a check in the fixture file that should help us detect when we fail to do so.

Anyway, take a look and let me know what you think. I do like it more than what I had.

Aaronius: Alternatively, we could just use a different test framework. 😆

PR #182, opened 2021-08-12
PDCL-6649 Fixed error sometimes being thrown when clicking Save in Launch.

Aaronius:
## Description
Copied from the JIRA issue since it explains it well:

Start to create a data element using the XDM Object data element type. Don't select any schema (you'll need to have more than one sandbox or more than one schema in order to not select a schema). Hit Save to Library (or one of the other Save buttons Launch provides). Notice it results in an error. You'll see "An error occurred" in the extension view.

When Save was clicked, the extension view was in an invalid state. Launch called the extension's validate function, which returned false, but then Launch also called getSettings. Our getSettings assumes that it will only be called if validate returns true. Because of this, the code ends up inadvertently throwing an error.

Now, in this scenario, Launch doesn't really need to be calling getSettings if validate returns false. I've logged that issue in https://jira.corp.adobe.com/browse/PDCL-6647. However, there's another scenario when getSettings may be called when the view is in an invalid state. When a user navigates away from the extension view without hitting the Save button, Launch requests settings from the extension and compares them to the persisted settings. If they are different, Launch assumes the extension view is dirty and asks the user if they would like to save or discard their changes before navigating. Launch doesn't call validate in this case.

For this reason, we need to place more guards inside the getSettings function so that it won't throw errors when it's called and the extension view is in an invalid state.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-6649

## Motivation and Context
Unexpected errors don't provide a good user experience.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/dataElements/xdmObject.jsx
+++ b/src/view/dataElements/xdmObject.jsx
@@ -89,6 +89,12 @@ const XdmObject = ({ initInfo, formikProps, registerImperativeFormApi }) => {
   useEffect(() => {
     registerImperativeFormApi({
       getSettings({ values }) {
+        if (!isEditorRenderable) {
+          // We're in an invalid state where we can't
+          // build a proper settings object.
+          return {};
+        }
+
         const schema = {
           id: selectedSchema.$id,
           version: selectedSchema.version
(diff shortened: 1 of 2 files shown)


jfkhoury (approved):
Does that mean when we reload the data element we won’t be able to pre-populate any inputs?

jonsnyder (approved):

Aaronius:
> Does that mean when we reload the data element we won’t be able to pre-populate any inputs?

No, everything else we're doing should be fine.

PR #185, opened 2021-08-19
PDCL-6598 Improved UX for editing item attributes within XDM Object data element.

Aaronius:
## Description
To demonstrate the UX issue, here are reproduction steps:

Here are the reproduction steps:

1. In the XDM Object data element view, click on an array field in the schema tree.
1. Click on "Add item" on the right hand side.
1. You should see "Item 1" now. Click "Item 1".

The right side of the view changes to show two radio buttons: "Provide individual attributes" and " Provide entire object". "Provide individual attributes" is selected, but it's not clear what the user is supposed to do at that point to actually provide the individual attributes. What we're expecting them to do is to click the attributes in the tree on the left and edit them that way, but that's just not clear to the user.

The solution provided in this PR consists of:

1. When select an array item in the array edit form OR select an item within the breadcrumb we display at the top of the form, we now automatically expand the corresponding node within the schema tree, and expand all its ancestor nodes, and scroll to the node so that it and all its child attributes are visible to the user.
2. On the object edit view, we now show a message that indicates the user can modify individual attributes by using the schema tree.
3. Whenever an item is selected by clicking a node in the schema tree, we expand the node to show its child fields. This part wasn't really necessary, but it did help make things consistent (that is, whenever an item gets selected through any means, the node in the tree gets expanded).

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-6598

## Motivation and Context
I put this in the description because it gave proper context.

## Types of changes

## Checklist:


src/view/components/typography/body.jsx
```diff
@@ -25,7 +25,14 @@ import getDimensionStyle from "../../utils/getDimensionStyle";
  * for more info. Once React-Spectrum provides a Body component
  * that handles typography, we can get rid of this.
  */
-const Body = ({ children, size = "S", isSerif, marginTop, marginBottom }) => {
+const Body = ({
+  "data-test-id": dataTestId,
```
Aaronius, line 29: These changes for adding support for `data-test-id` within `Body` and `Heading` are also in https://github.com/adobe/reactor-extension-alloy/pull/178.

src/view/dataElements/xdmObject/components/nodeEdit.jsx
```diff
@@ -78,6 +86,9 @@ const NodeEdit = props => {
           ))}
         </Breadcrumbs>
       </View>
+      <Heading data-test-id="heading" size="M">
+        {displayName}
+      </Heading>
```
Aaronius, line 91: Note that this heading is new and is showing the "display name". This wasn't really necessary, but I got carried away and included it anyway.

src/view/dataElements/xdmObject/helpers/getNodeEditData.js
```diff
@@ -75,14 +77,14 @@ const getNodeEditData = ({
         candidateFieldName: candidateFieldName
           ? `${candidateFieldName}.properties.${propertyName}`
           : `properties.${propertyName}`,
-        candidateDisplayName:
-          propertySchema.candidateDisplayName || propertyName
```
Aaronius, line 79: `propertySchema.candidateDisplayName` never was a thing. I think this was a mistake I made when I built the XDM Object view. I think what I meant was `candidateDisplayName: propertySchema.title || propertyName`. I fixed this, which caused the breadrumb to use the friendly display names (when available) and not the property names, and I thought it was a little awkward from a user perspective to have the text in the breadcrumb be inconsistent with the node labels shown in the schema tree.

So......I changed this code, so that we're using the "property name" in the breadcrumb, which matches what's shown in the schema tree. We now render the friendly display name as a heading on the edit form below the breadcrumb. I think it makes for a pretty good UX, but I'm open to feedback.

src/view/dataElements/xdmObject/components/nodeEdit.jsx
jonsnyder: I like it!

src/view/dataElements/xdmObject/components/objectEdit.jsx
```diff
@@ -57,6 +57,11 @@ const ObjectEdit = ({ fieldName }) => {
             width="size-5000"
           />
         </DataElementSelector>
+      ) : (
+        <div>
+          Please use the tree to the left to provide values for individual
+          attributes.
```
jonsnyder, line 63: What does "use the tree" mean?

Please click on an attribute in the tree on the left to provide values for individual attributes.

test/functional/helpers/clientScripts/preventSpecificErrorsFromFailingTests.js
```diff
@@ -47,4 +47,29 @@ window.addEventListener("error", event => {
...
+    // ignoring errors where event.message is "Script error." may also
+    // ignore unrelated errors that we DO actually want to fail tests,
+    // but our options are limited here.
+    event.message === "Script error."
+  ) {
+    // Using event.preventDefault here is sufficient to prevent the test
+    // from failing in Safari, but Firefox requires us to call
+    // event.stopImmediatePropagation instead.
+    event.stopImmediatePropagation();
+  }
```
Aaronius, line 74: I discussed this mess in Slack.

jonsnyder (approved):
The single breadcrumb and friendly name right next to each other look a little weird when selecting a top level attribute. Maybe we should only show bread crumbs when there are at least 2 levels?

Aaronius:
Thanks @jonsnyder. I've made the changes based on your feedback.

jonsnyder:
> Thanks @jonsnyder. I've made the changes based on your feedback.

Thanks. Looks good.

src/view/dataElements/xdmObject/helpers/getNodeEditData.js
jfkhoury: I like 👍🏻

jfkhoury (approved):
Awesome! Works cross browser?

PR #188, opened 2021-09-07
BuildInfo is deprecated in favor of environment.stage

jonsnyder:
Launch made a change which deprecated buildInfo. We need to know the environment to determine which data stream to use. The new property is in `turbine.environment.stage`.

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/instanceManager/createInstanceManager.js
+++ b/src/lib/instanceManager/createInstanceManager.js
@@ -35,9 +35,9 @@ module.exports = ({
       instanceByName[name] = instance;
 
       const computedEdgeConfigId =
-        (turbine.buildInfo.environment === "development" &&
+        (turbine.environment.stage === "development" &&
           developmentEdgeConfigId) ||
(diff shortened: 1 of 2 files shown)


src/lib/instanceManager/createInstanceManager.js
```diff
@@ -35,9 +35,9 @@ module.exports = ({
       instanceByName[name] = instance;
 
       const computedEdgeConfigId =
-        (turbine.buildInfo.environment === "development" &&
+        (turbine.environment.stage === "development" &&
```
jfkhoury, line 38: I think it's weird that they named it `stage` :)

jfkhoury (approved):

src/lib/instanceManager/createInstanceManager.js
jonsnyder: You and me both. I was thinking stage as in staging, but it is stage as in development stage.

PR #217, opened 2022-04-20
Improved Tags UI performance.

jonsnyder:
## Description

Upgrading the Parcel v2 reduced the size of the reactor package from 3.1M to 464K. Not only will this keep us from hitting the reactor package limit, but it should provide a much better experience using the extension. The UI will load much faster with these changes.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-8266

## Motivation and Context

Reactor/Tags has a package upload limit of 5MB. 

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/deployRelease.yml
+++ b/.github/workflows/deployRelease.yml
@@ -14,7 +14,7 @@ jobs:
     name: "Validate"
     runs-on: ubuntu-latest
     steps:
-      - uses: adobe/project-card-release-automation/validate-version@v1
+      - uses: adobe/project-card-release-automation/validate-version@v2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           version: ${{ github.event.inputs.version }}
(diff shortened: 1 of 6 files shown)


.babelrc
```diff
@@ -2,7 +2,6 @@
   "plugins": [
     "version"
   ],
-  "presets": [["@babel/preset-env"]],
```
jonsnyder, line 5: Parcel includes its own babel plugins, and produces larger files if this is left in. For the parts of the build using Babel standalone, I add this back in via a command line argument.

src/view/components/codePreview.styl
```diff
@@ -16,7 +16,7 @@ governing permissions and limitations under the License.
   white-space: pre;
 }
 
-.CodePreview-openEditorButton {
+button.CodePreview-openEditorButton {
```
jonsnyder, line 19: Parcel v2 changed the order of some of the css files so that this was not overriding the spectrum style. Adding button makes this more specific so that it will override.

test/functional/configuration/configuration.spec.js
```diff
@@ -286,6 +286,8 @@ test("returns full valid settings", async () => {
   await instances[1].idMigrationEnabled.click();
   await instances[1].thirdPartyCookiesEnabled.click();
   await instances[1].onBeforeEventSendEditButton.click();
+  // Click on the field before clearing to get rid of the "..."
+  await instances[1].downloadLinkQualifierField.click();
```
jonsnyder, line 290: I suspect moving to parcel v2 introduced a testCafe timing issue when clearing long textfields.

carterworks (approved):

PR #235, opened 2022-05-31
Added redirect with identity action to allow cross-domain identity sharing.

jonsnyder:
## Description

This adds an action that allows you to append the identity to urls and redirect to the new url. This is designed to work within a rule with a Core - click event. The click event should have a selector of 'a[href]'. Then you can filter links using conditions with %this.hostname% data element. Then the 'Redirect with identity' action will append the identity to the URL and redirect.

## Types of changes

## Checklist:


scripts/functional.js
```diff
@@ -67,25 +61,9 @@ const buildComponentFixtures = async () => {
   return bundler.run();
 };
 
-const buildRuntimeFixtures = async () => {
```
jonsnyder, line 70: I decided to just use the alloyio.com functional test endpoint instead of building a new fixture. This allowed me to test running against multiple domains.

jfkhoury: Awesome decision.

src/lib/actions/redirectWithIdentity/createRedirectWithIdentity.js
```diff
@@ -0,0 +1,51 @@
...
+      `Invalid event target when running "Redirect with identity." This action is meant to be used with a Core click event using an "a[href]" selector.`
+    );
+    return Promise.resolve();
+  }
+
+  if (event.nativeEvent.preventDefault) {
+    event.nativeEvent.preventDefault();
+  }
+
+  const url = event.nativeEvent.target.href;
```
jfkhoury, line 45: Do we need to filter invalid `hrefs` or no need? You know how people do `href=#`...

jfkhoury: I guess customers can do that in their `Condition` right?

test/functional/runtime/redirectWithIdentity.spec.js
```diff
@@ -0,0 +1,127 @@
...
+import { t, ClientFunction, Selector } from "testcafe";
+import createRuntimeFixture from "./helpers/createRuntimeFixture";
+import createNetworkLogger from "./helpers/createNetworkLogger";
+import addHtmlToBody from "./helpers/addHtmlToBody";
+import { SECONDARY_TEST_PAGE } from "./helpers/constants/url";
+import getReturnedEcid from "./helpers/getReturnedEcid";
+
+const networkLogger = createNetworkLogger();
+
+const container = {
```
jfkhoury, line 22: I wonder if we need to abstract this container and replace it with a creator function?

It seems that huge chunks of it will be the same across tests. What do you think J?

jfkhoury: It makes it easier to use, especially for other engs.

Something similar to the `createRuntimeFixture` helper.

carterworks: I agree. Something like `createContainerSettings(overrides)`, which returns `{...defaultContainer, ...overrides}`.

src/lib/actions/redirectWithIdentity/createRedirectWithIdentity.js
carterworks: And will this work with relative URLs?

jonsnyder: It will work with relative URLs or hashes. The customer should filter these out though because they would be for the same domain.

test/functional/runtime/redirectWithIdentity.spec.js
jonsnyder: Yes I agree.

carterworks (changes requested):
[Add the function to construct the basic container](https://github.com/adobe/reactor-extension-alloy/pull/235/files/9c9e6b9b984528d1b0b60f3ad98e352daeb35054#r889185630)
and we will be good to go.

ninaceban (approved):
Looks good to me! 
In case there is only one instance configured, should we show it pre-selected and disabled?

jonsnyder:
I'm going to merge now without creating a container builder. We can do that later.

jonsnyder:
> Looks good to me!
> In case there is only one instance configured, should we show it pre-selected and disabled?

Probably a good idea.

PR #262, opened 2022-10-06
Fixed bug with identityMap having incorrect placement on xdm

dcottingham:
## Description

https://jira.corp.adobe.com/browse/TNT-45867

While troubleshooting a customer complaint about target3rdpartyId not working, I came across a bug in the identityMap data element.  Currently the identityMap is added directly to the xdm object, which is incorrect and results in Web SDK leaving it off the interact call. i.e. `{
   "xdm":{
      "target3rdpartyId":[
         {
            "id":"abcd",
            "authenticatedState":"authenticated"
         }
      ]
   }
}`

According to [this customer doc](https://experienceleague.adobe.com/docs/experience-platform/edge/personalization/adobe-target/using-mbox-3rdpartyid.html?lang=en#), the correct format should be: 
`{
   "xdm":{
      "identityMap":{
         "target3rdpartyId":[
            {
               "id":"abcd",
               "authenticatedState":"authenticated"
            }
         ]
      }
   }
}`

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/dataElements/identityMap/createIdentityMap.js
+++ b/src/lib/dataElements/identityMap/createIdentityMap.js
@@ -18,26 +18,33 @@ governing permissions and limitations under the License.
 module.exports = ({ logger }) => {
   return settings => {
     // settings _are_ the identity map
-    return Object.keys(settings).reduce((newIdentityMap, namespace) => {
-      const filteredIdentifiers = settings[namespace].filter(({ id }, i) => {
-        const isValidId = typeof id === "string" && id.length;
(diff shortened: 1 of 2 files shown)


carterworks (approved):

jonsnyder (commented):
Don, thanks for taking a look at this issue. I think the customer is using the extension incorrectly. I'm guessing they put the identityMap data element as the XDM object directly. What they need to do is define an XDM object data element with the identityMap field set to their identityMap data element. Then use the XDM object data element in the send event action.

PR #301, opened 2023-04-19
Revert "Added decisioning event types to support top of page hits."

carterworks:
Reverts adobe/reactor-extension-alloy#299


carterworks:
We decided to name it `propositionFetch`

PR #307, opened 2023-04-24
Deprecate the datasetId option

carterworks:
## Description

The following changes are on the `sendEvent` action configuration page

- Add a deprecation notice to the label and field description for the `datasetId` option
- When loading existing settings, take the old `datasetId` value and put it into `edgeConfigOverrides.com_adobe_experience_platform.datasets.event.datasetId`
- When saving new settings, take anything saved in `datasetId` and put it in `edgeConfigOverrides.com_adobe_experience_platform.datasets.event.datasetId` instead. This codepath will never be run, but it's a safety measure.
- Keep the `datasetId` input field, but have it mirror exactly the `edgeConfigOverrides.com_adobe_experience_platform.datasets.event.datasetId` input field. They map to the same value and changes in one reflect the other.
- Create `deepAssign()` (with unit tests), a small utility function that assigns to a string path of an nested object

For example,

```js
const obj = {};
deepAssign(obj, "a.b.c", 1);
console.log(obj); // { a: { b: { c: 1 } } }
```

## Related Issue

[PDCL-10349: Deprecate `datasetId` option in Web SDK Tags Extension UI](https://jira.corp.adobe.com/browse/PDCL-10349)

## Motivation and Context

With the advent of datastream configuration overrides, [we have deprecated the `datasetId` option for the `sendEvent` command. Alloy shows a console warning to use the new option when `datasetId` is used](https://github.com/adobe/alloy/blob/08c77750e9f9db23fbff84d744f3e300888ab181/src/components/DataCollector/index.js#L71). 

image

## Types of changes

## Checklist:

### DIFF
--- a/src/view/actions/sendEvent.jsx
+++ b/src/view/actions/sendEvent.jsx
@@ -28,6 +28,7 @@ import Overrides, { bridge as overridesBridge } from "../components/overrides";
 import { DATA_ELEMENT_REQUIRED } from "../constants/validationErrorMessages";
 import FormElementContainer from "../components/formElementContainer";
 import InstanceNamePicker from "../components/instanceNamePicker";
+import { deepAssign } from "../utils/deepAssign";
 
 const getInitialValues = ({ initInfo }) => {
   const {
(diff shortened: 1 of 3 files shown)


carterworks:
Instead of risking bugs right as we release with the marshaling/migrating around of old options, we are going to just change the label. Phase 2 is to use extension transforms.

PR #315, opened 2023-05-15
Enhanced the config overrides UI

carterworks:
## Description

Call this "config overrides UI v2". v1 was a simple list of text fields. v2 is a rich interface that helps the user select the right value.

- Require that overrides are created on a per-environment basis (i.e. production, staging, and development have separate overrides).
- Pull the existing Blackbird configurations for potential overrides and display them in a dropdown.
- Allow entry of report suite overrides as a comma-separated list.
- Add a link to the documentation.
- reload the UI and refetch options if there is a change in the orgID/datastream ID settings
- If there is a dropdown present, validate that the typed options are in the dropdown (but ultimately still allow them to put in their custom value).
- Filter out blank entries and report suites.

## Related Issue

[PDCL-10323: Add dropdowns to Config Overrides in Tags Extension](https://jira.corp.adobe.com/browse/PDCL-10323)

image

## Types of changes

## Checklist:


src/lib/actions/redirectWithIdentity/createRedirectWithIdentity.js
```diff
@@ -10,11 +10,13 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-module.exports = ({ instanceManager, document, logger }) => (
+const createConfigOverrides = require("../../utils/createConfigOverrides");
```
jonsnyder, line 13: Any time you require something in a file, it can make unit testing harder. In this case, in order to test that the correct overrides are used, you end up testing the logic in createConfigOverrides too. The tests stop being unit tests. For example, if there is a bug in createConfigOverrides, then a bunch of the tests are going to fail. There is a balance though because javascript's built in functions are quite limited. So sometimes it is okay to require a utility.

So what is properly a utility? Is it anything that is in the "utils" folder? Maybe. Certainly it needs to be a pure function, and createConfigOverrides is a pure function. I think it's important also that a util is something that has been thoroughly tested. I think what fails here though is that a utility should be something that is generally applicable, and not something that implements specific business logic. (i.e. a map function, a find, or assign).

Another way I like to think about it is that if we are following dependency injection principles, we should be able to separate the code into units that create new instances of objects and units that don't. The standard we've come to within the Alloy repos is that "index.js" files can create object/functions, and import/require other files, while the rest of the code tries to stay within it's unit.

So here is a different approach. Change the signature of createConfigOverrides to this:
```
const createGetConfigOverrides = environmentName => settings => {
```
Then in your index.js files do this:
```
const createRedirectWithIdentity = require("./createRedirectWithIdentity");
const instanceManager = require("../../instanceManager/index");
const createGetConfigOverrides = require("../../utils/createGetConfigOverrides");

const getConfigOverrides = createGetConfigOverrides(turbine.environment.stage);
const document = window.document;
module.exports = createRedirectWithIdentity({
   instanceManager,
   document,
   logger: turbine.logger,
   getConfigOverrides
});
```
So now you are passing a function to extract config overrides out of a settings object instead of having to pass the whole turbine object and having to require createConfigOverrides. Now when you run unit tests on createRedirectWithIdentity, you only have to verify that the getConfigOverrides function gets all the settings, and that whatever is returned from getConfigOverrides is sent to redirectWithIdentity.

jonsnyder (changes requested):
I haven't had a chance to review the whole PR, but I wanted to submit this so that you can start working on it. Let me know if you have questions about this, I can talk you through it.

jonsnyder (changes requested):
I was able to run the PR today. I've got a few notes. Maybe we can discuss in standup or something:

1. In the tabs, I would expect them to be ordered "Development", "Staging", "Production". The same way they are ordered visually when moving libraries to each stage in Tags.
2. I love the visual separation in the configuration view, but there isn't a similar header in the actions. It looks a bit busy and hard to follow in the actions. I'm leaning towards using the header from configuration, but is that weird because there are no other headers?
3. I noticed that when the settings are saved for report suites, you save empty rows. In other places we just filter out empty rows (like in the decision scopes UI).
4. Do we want to add a warning if you'd typed in an ID that isn't configured as a secondary ID? We wouldn't want the warning if you specified a data element though.
5. The copy to functionality is a bit involved. I think it would be better to just have two buttons. Also when thinking about it further, I think it makes more sense to put the buttons at the top of the tabs and reverse the functionality. So on the Production tab, you'd have "Duplicate staging settings" and "Duplicate development settings". These buttons would copy the staging or dev settings and put them in the current view. The way you have it, it is difficult to know if the copy button did anything because you don't see anything change. Also I think the button should be styled similar to the "Restore default" buttons we have.
6. I'm not sure the "i" buttons fit into the spectrum style. Should we put this information in the description text? Also I think a better wording is "Event dataset from datastream:" If we do keep the information popups, you should use a labeledValue component from react spectrum.
7. We have this concept of when you don't specify a staging or dev datastream id, we just use the production one. This was carried over from the AppMeasurement extension. Do we want to apply that same principle to the config overrides? If no staging datastream is specified, should we use the production overrides in a staging environment?
8. On the set consent action, the config overrides should be at the bottom.

jonsnyder (commented):
By the way, awesome job! These UIs are very difficult especially when you add network calls. I wasn't able to find any bugs with the functionality.

jonsnyder:
For 7. I just wanted to have a conversation about it. I am advocating something else.
* Only show the config override tabs for the environments with data streams that have been specified. With current functionality we fallback to the production datastream when dev and/or staging is not configured. Now that I'm thinking through all the edge cases, maybe this doesn't make sense.

carterworks:
Let me know what you think 

Previous small header on `sendEvent` action configuration page.

image

New large header that matches the extension configuration page.

image

carterworks:
I've moved the "what value am I overriding" into the field description. Here is how it looks.
image

src/lib/actions/redirectWithIdentity/index.js
```diff
@@ -12,11 +12,13 @@ governing permissions and limitations under the License.
...
+const createGetConfigOverrides = require("../../utils/createGetConfigOverrides");
 
 const document = window.document;
 
 module.exports = createRedirectWithIdentity({
   instanceManager,
   document,
-  logger: turbine.logger
+  logger: turbine.logger,
+  getConfigOverrides: createGetConfigOverrides(turbine.environment?.stage)
```
ninaceban, line 23: Why are we passing the stage environment value?

carterworks: We need the stage (`development`, `production`, or `stage`) because overrides are now configure on a per-environment basis and we need to know which ones to get.

PR #321, opened 2023-06-19
Added support for per-command overrides of the datastream ID

carterworks:
## Description

The Tags companion of [PR #995 in Alloy](https://github.com/adobe/alloy/pull/995). Needs to be merged after #315 because it is based on that PR.

Adds two new dropdowns to the overrides section of the actions "send event", "set consent", and "redirect with identity": "sandbox" and "datastreamId". This means that there are a number of new settings saved with the extension: a `datastreamId` and a `sandbox` for every environment. 

When the action is first loaded, it will request to get all available sandboxes. If there is only one, it will default to that value and disable the dropdown. Once a sandbox is selected, it will request all the available datastreams and display them in a picker. Once a datastream is selected, the form will request the blackbird configuration for the other fields. If at any point there is an error or no results are shown, the form will fall back to manual entry. The user can also select, at any time, to input a value manually or specify a data element.

image

## Related Issue

[PDCL-10571: Allow for overriding the datastreamId in Alloy](https://jira.corp.adobe.com/browse/PDCL-10571)

## Types of changes

## Checklist:


src/view/components/formikReactSpectrum3/formikComboBox.jsx
```diff
@@ -15,7 +15,21 @@ import PropTypes from "prop-types";
...
+ * @param {string?} params.width
+ * @param {((val: string) => boolean)?} params.isValid A function that is called with
+ * the value of the input and expects boolean to be returned representing the
+ * validity of the input.
+ * @returns {React.Element}
+ */
+const FormikComboBox = ({
+  name,
+  width,
+  isValid = () => true,
```
jonsnyder, line 30: Why is this needed? If the Yup validation fails then this would show an error anyways.

src/view/components/overrides/overrides.jsx
```diff
@@ -9,187 +9,379 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
+                          width="size-5000"
+                        />
+                        <DatastreamSelector
+                          data-test-id={FIELD_NAMES.datastreamId}
+                          label="Datastream"
+                          description={`Override the default datastream${
+                            envEdgeConfigIds.datastreamId
+                              ? ` (${envEdgeConfigIds.datastreamId})`
+                              : ""
+                          }. Does not support cross-organiztion datastream overrides.`}
```
jonsnyder, line 269: "Override the configured datastream ${...}. This does not support cross-organization datastream overrides."

extension.json
```diff
@@ -117,57 +117,178 @@
...
-                              "datasetId": {
-                                "type": "string"
-                              }
-                            },
-                            "required": ["datasetId"]
+                  "development": {
+                    "datastreamId": {
+                      "type": "string",
+                      "minLength": 1
+                    },
```
jonsnyder, line 126: Where is sandbox?

extension.json
```diff
@@ -194,18 +315,6 @@
...
-        "type": "remove",
-        "propertyPath": "instances[].sandbox"
-      },
-      {
-        "type": "remove",
-        "propertyPath": "instances[].stagingSandbox"
-      },
-      {
-        "type": "remove",
-        "propertyPath": "instances[].developmentSandbox"
```
jonsnyder, line 208: Why get rid of these?

jonsnyder (commented):
Looks good. We should do a demo sometime with Mitch. 

- The spacing above the section header for configuration overrides is inconsistent. It should be a large space like it is in the extension configuration view for the other sections.
- If you set a datastream override, the other override fields do not show drop downs.

extension.json
carterworks: I deleted them while trying something out and forgot to add them back.

carterworks:
> If you set a datastream override, the other override fields do not show drop downs.

It will if the selected datastream has overrides defined in the launch config. Otherwise, it will fallback to just a text input.

test/functional/actions/sendEvent.spec.js
```diff
@@ -383,3 +659,85 @@ test("shows error for data value that is more than one data element", async () =
...
+  });
+
+  await overrideViewSelectors.envTabs.production.click();
+  await overrideViewSelectors.envTabs.production.expectSelected();
+  await overrideViewSelectors.datastreamInputMethod.freeform.click();
+  await overrideViewSelectors.datastreamInputMethod.freeform.expectChecked();
+  await overrideViewSelectors.datastreamIdFreeform.typeText("1234");
+
+  await extensionViewController.expectIsValid();
+});
```
jonsnyder, line 743: Can you add a test here that when choosing another instance, the datastream overrides switch to textfields and vice versa?
Also add a test that when you override the datastream, the drop downs populate with the allow lists of that datastream.

test/functional/configuration/configuration.spec.js
```diff
@@ -1193,3 +1333,294 @@ test.requestHooks(
...
+  await t.pressKey("tab");
+  await instances[0].overrides.comboBoxes.targetPropertyTokenOverride.expectNoError();
+
+  await instances[0].overrides.comboBoxes.reportSuiteOverrides[0].expectExists();
+  await instances[0].overrides.comboBoxes.reportSuiteOverrides[0].enterSearch(
+    "unifiedjsqeonly2, %Alloy Data Element%"
+  );
+  await t.pressKey("tab");
+  await instances[0].overrides.comboBoxes.reportSuiteOverrides[0].expectNoError();
+});
```
jonsnyder, line 1626: Can you add a test that for the second instance it shows free-form fields?

jonsnyder (approved):
I just looked at the code. I'd love to be a part of a manual testing session.

extension.json
```diff
@@ -117,57 +117,202 @@
...
+                      "required": ["propertyToken"]
+                    }
                   },
-                  "com_adobe_identity": {
-                    "type": "object",
-                    "properties": {
-                      "idSyncContainerId": {
-                        "type": "integer"
-                      }
+                  "staging": {
```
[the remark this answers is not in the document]
carterworks: It would make things a lot easier, but

1. Each of the definitions is technically inside a different json schema and you can't do `$ref` across schemas
2. Even if we could do cross-schema `$ref`, [the schema for `extension.json` doesn't allow for new top-level keys like `$defs`](https://github.com/adobe/reactor-turbine-schemas/blob/master/schemas/extension-package-web.json#L70).

The solution would be to create a script that generates our `extension.json` and have that run as part of our `dev` and `build` scripts. I can do that if you think it's not too much out of scope.

PR #327, opened 2023-07-20
Fix config overrides issues

carterworks:
## Description

- [x] When you are upgrading from an existing installation, Launch blocks it because the old settings aren't valid in the new extension schema.
- [x] When an input fails validation (like if you put in a random string into a combobox, but the string isn't in the list), it does not display why it failed
- [x] If you put a random value in one environment, then use the button to copy, they don't validate again because they aren't marked as dirty.
- [x] Cannot unselect a datastream override from a list.
- [x] Partial data elements (`1234%data element%`) are not supported in the id sync container
- [x] Give the id sync container validation a friendly error message
- [x] The datastream selector does not reload when selecting a different sandbox
- [x] This error (see image) that Mitch Rice encountered    

image

## Types of changes

## Checklist:

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -119,6 +119,57 @@
                 "type": "object",
                 "additionalProperties": false,
                 "properties": {
+                  "com_adobe_experience_platform": {
+                    "type": "object",
+                    "properties": {
+                      "datasets": {
+                        "type": "object",
+                        "properties": {
+                          "event": {
+                            "type": "object",
+                            "properties": {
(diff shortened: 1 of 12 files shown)


carterworks:
>  If you put a random value in one environment, then use the button to copy, they don't validate again because they aren't marked as dirty.

I have been trying to resolve this issue all day. No combination of `setTouched` with field names seems to work, so I am going to skip it for foreseeable future.

PR #343, opened 2023-08-07
Fixed issue where data elements could not be saved in ID sync container ID field

carterworks:
## Description

Where the `extension.json` previously only allowed `number` in the ID sync container ID field, with this change, it allows either a `number` or a `string` that matches the regex `^([^%\n]*%[^%\n]+%)+[^%\n]*$` for data elements.

## Related Issue

[PLATIR-30215](https://jira.corp.adobe.com/browse/PLATIR-30215?focusedCommentId=37547998&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-37547998)

## Motivation and Context

Our automated tests do not test the `extension.json`, so it slipped through the cracks.

## Types of changes

## Checklist:

### DIFF
--- a/scripts/helpers/createExtensionManifest.js
+++ b/scripts/helpers/createExtensionManifest.js
@@ -77,7 +77,15 @@ const createEdgeConfigOverridesSchema = isAction => {
       type: "object",
       properties: {
         idSyncContainerId: {
-          type: "integer"
+          anyOf: [
+            {
+              type: "integer"
+            },
+            {
+              type: "string",
+              pattern: "^([^%\n]*%[^%\n]+%)+[^%\n]*$"
+            }
+          ]
         }
       },
       required: ["idSyncContainerId"]
--- a/scripts/helpers/createExtensionManifest.js
+++ b/scripts/helpers/createExtensionManifest.js
@@ -82,8 +82,7 @@ const createEdgeConfigOverridesSchema = isAction => {
               type: "integer"
             },
             {
-              type: "string",
-              pattern: "^([^%\n]*%[^%\n]+%)+[^%\n]*$"
+              type: "string"
             }
           ]
         }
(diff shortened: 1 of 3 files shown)


carterworks:
I've created [PDCL-10847](https://jira.corp.adobe.com/browse/PDCL-10847) to run the validation from `extension.json` in our functional tests, which would have caught this issue.

jonsnyder (approved):
You could just set it to a string. I don't know what value this regex adds. Or you could change this regex to only allow [0-9] instead of [^%\n]. For example your reportSuite validation is just a string.

Do we need to do a parseInt call in the runtime so that it is passed as an integer into experience edge?

carterworks:
> You could just set it to a string. I don't know what value this regex adds.

That's fair. Excess validation has burned me a lot recently, specifically with this field.

> Do we need to do a parseInt call in the runtime so that it is passed as an integer into experience edge?

[We do that before saving the settings into Tags](https://github.com/adobe/reactor-extension-alloy/blob/main/src/view/components/overrides/overridesBridge.js#L124).

carterworks:
@jonsnyder said

> If the customer wants to use a data element for the containerId, it will be saved as a string in the settings. When the tags library is run, Tags will resolve the data element and the containerId will be a string containing an integer.

I just pushed a fix that looks if the ID sync container ID is a string at runtime and runs `parseInt()` on it, if it is.

jonsnyder (approved):

ninaceban (commented):
why we return `undefined` at line 35? 
```
// there are no settings for this current env, but there are settings for others
    if (
      edgeConfigOverrides.development ||
      edgeConfigOverrides.staging ||
      edgeConfigOverrides.production
    ) {
      return undefined;
    }
```

carterworks:
> why we return `undefined` at line 35? 
> ```
> // there are no settings for this current env, but there are settings for others
>     if (
>       edgeConfigOverrides.development ||
>       edgeConfigOverrides.staging ||
>       edgeConfigOverrides.production
>     ) {
>       return undefined;
>     }
> ```

We are returning `undefined` because there are no overrides for the current environment. It is `return undefined` instead of `return` because there is an eslint rule where functions that return values should always return a value.

ninaceban (approved):

PR #375, opened 2024-01-19
Update dependencies to allow project being used with node 20/21.

dompuiu:
## Description

I was not able to install the project dependencies when I was using node v20 or v21 on M1 Max. When using node v18, everything worked .

The issue seems was coming form a `parcel` dependency that was locked in the `package-lock.json`. I updated `parcel` to the latest versions and deleted `package-lock.json` before doing a fresh install. 

## Motivation and Context

Node v18 is in maintenance and the current LTS version is v20.

## Types of changes

## Checklist:


jonsnyder:
Did this break anything? Why did you close?

dompuiu:
I got to a point where the functional tests were not running. But at that time I wasn't able to run the tests on the main branch either. I didn't want to waste more time before finishing the other stories. 

And probably I will split this into multiple chunks. Prettier 3 has changed its default values. Only when I upgrade prettier, many files will be changed by the `lint` command. It will become hard to follow if all the dependencies are updated in one PR.

I will return to this once I finish the stories and I will upgrade the dependencies in chunks.

dompuiu:
Another thing. Some of the latest packages are no longer exporting CJS modules (just ES). Because of that, a lot of scripts need to be changed to `.mjs` files. So it adds up.

PR #401, opened 2024-05-24
subscribeContentCards event

jasonwaters:
## Description

This PR adds a new event type to support the `subscribeContentCards` command.  See related alloy PR: https://github.com/adobe/alloy/pull/1125

This is my first time contributing to this repo, so please take a close look and let me know your feedback! 😄 

screenshot-2024-05-24 at 16 50 22@2x

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/scripts/helpers/createExtensionManifest.mjs
+++ b/scripts/helpers/createExtensionManifest.mjs
@@ -963,6 +963,36 @@ const createExtensionManifest = ({ version }) => {
             ]
           }
         }
+      },
+      {
+        name: "subscribe-content-cards",
+        displayName: "Subscribe content cards",
+        libPath: "dist/lib/events/subscribeContentCards/index.js",
+        viewPath: "events/subscribeContentCards.html",
+        schema: {
+          $schema: "http://json-schema.org/draft-04/schema#",
+          type: "object",
+          instanceName: {
+            type: "string",
+            minLength: 1
+          },
+          surfaces: {
+            anyOf: [
+              {
+                type: "array",
+                minItems: 1,
+                items: {
+                  type: "string",
+                  minLength: 1
+                }
+              },
+              {
+                type: "string",
+                pattern: "^%[^%]+%$"
+              }
+            ]
+          }
+        }
       }
     ],
     dataElements: [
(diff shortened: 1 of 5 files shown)


jfkhoury:
@jasonwaters Do you think we need a phase 2 here where we can provide move utils to the users to avoid custom code?

jasonwaters:
@jfkhoury Maybe!  I removed the insanely convenient `subscribeContentCards` command at the direction of the alloy team.  In the spirit of that I think we are going to take a wait-and-see approach to see how much support customers want here, or if `subscribeRulesetItems` is sufficient and offers enough functionality and flexibility without too much agony.

ninaceban:
@jasonwaters thank you! In one of your examples for Content Cards, you are sorting the propositions, do you think it will be useful to add sorting functionality to all types of items, or that is only applicable to the Content Cards?

ninaceban (approved):

ninaceban:
if `Message-Feed` item was deprecated, can we remove it from the drop-down, or mark it as `deprecated`, or just rename to `Content Cards` if it was just renamed?

jasonwaters:
@ninaceban thanks for calling that out, i didn't realize message feed item was previously added.  I removed it so that content card is there but not message feed item. 🙂

ninaceban:
> @ninaceban thanks for calling that out, i didn't realize message feed item was previously added. I removed it so that content card is there but not message feed item. 🙂

Do you know if we have customers using message-feed already?
I would suggest to have a logic when initializing the view that if the settings contain message-feed schema, we show content cards and we save content cards in the settings. This way we move the existing implementations to new schema and it won't throw errors.

jasonwaters:
@ninaceban content cards hasn't been released yet. That's what this PR is for. 🤪

The message feed schema was inadvertently released alongside in app messages. But no customers are using it, since there's no implementation or use case prior to this PR.

ninaceban:
> @ninaceban content cards hasn't been released yet. That's what this PR is for. 🤪
> 
> The message feed schema was inadvertently released alongside in app messages. But no customers are using it, since there's no implementation or use case prior to this PR.

I know :) I was worried about existing implementations that are using `message-feed` schema, because if we remove it, when they will open the view the dropdown will throw an error.  

If you think we are safe to just remove it. I'm ok with that.

jasonwaters:
@jonsnyder this needs a second 👍 to merge.

jonsnyder (approved):

PR #405, opened 2024-05-30
Fix a crash of the report suites override section.

dompuiu:
## Description

Fix a crash of the report suites override section. Niklas from Scania reported this on the Measure slack channel.

## Types of changes

## Checklist:

### DIFF
--- a/scripts/helpers/createExtensionManifest.mjs
+++ b/scripts/helpers/createExtensionManifest.mjs
@@ -327,6 +327,8 @@ const createExtensionManifest = ({ version }) => {
                       type: "integer",
                     },
                   },
+                  required: ["channel", "playerName"],
+                  additionalProperties: false,
                 },
                 personalizationStorageEnabled: {
                   type: "boolean",
@@ -1172,7 +1174,12 @@ const createExtensionManifest = ({ version }) => {
                   type: "array",
                   minItems: 1,
                   items: {
-                    enum: ["analytics", "target", "audiencemanager", "audienceManager"]
+                    enum: [
+                      "analytics",
+                      "target",
+                      "audiencemanager",
+                      "audienceManager",
+                    ],
                   },
                   required: ["name"],
                   additionalProperties: false,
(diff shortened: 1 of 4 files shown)


jfkhoury (approved):
👍🏻. Can we add a test for this? And can we proactively look for similar possible issues in this same workflow?

dompuiu:
> 👍🏻. Can we add a test for this? And can we proactively look for similar possible issues in this same workflow?

I was thinking about adding a test. But we wouldn't test any workflow. I am not sure if it is worth it. I tested all the other fields in the section. Report suites failed, because you can define more than one.

scripts/helpers/createExtensionManifest.mjs
```diff
@@ -327,6 +327,8 @@ const createExtensionManifest = ({ version }) => {
                       type: "integer",
                     },
                   },
+                  required: ["channel", "playerName"],
```
dompuiu, line 330: Schema was to relaxed. You could have saved an object without channel or playerName. You could have saved also other properties that were not defined.

src/view/configuration/streamingMediaSection.jsx
```diff
@@ -12,38 +12,58 @@ governing permissions and limitations under the License.
...
+
 export const bridge = {
   getInstanceDefaults: () => ({
     streamingMedia: getDefaultSettings(),
   }),
+
   getInitialInstanceValues: ({ instanceSettings }) => {
     if (!instanceSettings.streamingMedia) {
-      instanceSettings.streamingMedia = getDefaultSettings();
+      return bridge.getInstanceDefaults();
```
dompuiu, line 38: The extension configuration was crashing because the instanceSettings was returned. The instance settings was containing keys that could potentially override other bridge method from different sections. We need to return an object that contains just the `streamingMedia` key.

src/view/configuration/streamingMediaSection.jsx
```diff
@@ -12,38 +12,58 @@ governing permissions and limitations under the License.
...
+
+    return { streamingMedia };
   },
+
   getInstanceSettings: ({ instanceValues }) => {
     const instanceSettings = {};
     const { streamingMedia } = instanceValues;
 
+    ["appVersion", "adPingInterval", "mainPingInterval"].forEach((key) => {
+      if (!streamingMedia[key]) {
```
dompuiu, line 62: Here I clear the empty values of optional fields. Before, they were saved to the DB.

jonsnyder (approved):

PR #425, opened 2024-08-09
Added monitoring hook triggered event

ninaceban:
## Types of changes

## Checklist:

### DIFF
--- a/test/unit/lib/instanceManager/createInstanceManager.spec.js
+++ b/test/unit/lib/instanceManager/createInstanceManager.spec.js
@@ -92,6 +92,43 @@ describe("Instance Manager", () => {
       ],
     });
   });
+  it("adds a new monitor to the window.__alloyMonitors", () => {
+    const monitor = {
+      "onInstanceCreated": jasmine.createSpy("onInstanceCreated")
+    };
+    build();
+    expect(mockWindow).toEqual({
+      alloy1,
+      alloy2,
+      __alloyNS: ["alloy1", "alloy2"],
+      __alloyMonitors: [
+        {
+          onInstanceCreated: jasmine.any(Function),
+          onInstanceConfigured: jasmine.any(Function),
+          onBeforeCommand: jasmine.any(Function),
+        },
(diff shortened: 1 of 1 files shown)


src/lib/instanceManager/createInstanceManager.js
```diff
@@ -22,6 +22,33 @@ module.exports = ({
   const { instances: instancesSettings } = turbine.getExtensionSettings();
   const instanceByName = {};
 
+  const calledMonitors = {};
```
jonsnyder, line 25: This object represents a memory leak. I wasn't worried before when we were talking about just instanceCreated, and instanceConfigured, but now with more methods that will get called as the page progresses, this could represent a memory leak especially if someone called evaluateRulesets on a regular cadence. Maybe after 10 seconds we should remove the monitor that just records to calledMonitors. I was looking for an event in Tags that would be called when everything is initialized, but I couldn't find one.

jonsnyder (approved):

jonsnyder (approved):
👍

src/lib/instanceManager/createInstanceManager.js
```diff
@@ -101,12 +101,12 @@ module.exports = ({
...
-        calledMonitors[methodName].forEach((args) => {
-          if (newMonitor[methodName]) {
+        if (newMonitor[methodName]) {
+          calledMonitors[methodName].forEach((args) => {
             newMonitor[methodName](...args);
-            calledMonitors[methodName] = [];
-          }
-        });
+          });
+          calledMonitors[methodName] = [];
```
jonsnyder, line 108: Why blank out the calledMonitors? This makes it so that if you have two events with onConfigure hooks, only one of them will be called.

carterworks (approved):
I echo what Jon said.

src/lib/instanceManager/createInstanceManager.js
ninaceban: yeah, fixed it.

PR #468, opened 2025-02-26
Added graceful handling of missing data elements in update variable actions

jonsnyder:
## Description

Previously, if you deleted a data element, all your update variable actions with that data element would just show an error and not let you edit. Also if you copy a property all the update variable actions in the new property would still reference the data elements in the copied property and would just show the error too. This change catches the data element not found error and allows you to select a new data element, filling in all the values from before.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/actions/updateVariable.jsx
+++ b/src/view/actions/updateVariable.jsx
@@ -47,6 +47,7 @@ import {
   ADOBE_AUDIENCE_MANAGER,
   ADOBE_TARGET,
 } from "../constants/solutions";
+import deepAssign from "../utils/deepAssign";
 
 const isDataVariable = (data) => data?.settings?.solutions?.length > 0;
 
@@ -60,6 +61,10 @@ const getInitialFormStateFromDataElement = async ({
   existingFormStateNode,
   signal,
 }) => {
+  const value = {};
+  deepAssign(value, context.originalData);
+  deepAssign(value, data);
+
   if (
     dataElement.settings &&
     dataElement.settings.schema &&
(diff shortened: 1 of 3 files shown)


src/view/actions/updateVariable.jsx
```diff
@@ -60,6 +61,10 @@ const getInitialFormStateFromDataElement = async ({
   existingFormStateNode,
   signal,
 }) => {
+  const value = {};
+  deepAssign(value, context.originalData);
+  deepAssign(value, data);
```
carterworks, line 66: Why use `deepAssign` over `Object.assign` or `structuredClone`?

carterworks (approved):

src/view/actions/updateVariable.jsx
jonsnyder: Object.assign isn't recursive. Value could be a deeply nested object. I suppose the first deepAssign call could use structuredClone.

carterworks: Since we are using `=` to copy the things from the old object to the new one, the new one is full of references to the old object. So, for example

```js
const original = { a: { b: "alloy" } };
const clone = deepAssign({}, original);
original.a.b = "adobe";
console.log(clone.a.b); // "adobe"
```
Even though we didn't touch the clone.
If we do
```js
const original = { a: { b: "alloy" } };
const clone = structuredClone(original);
original.a.b = "adobe";
console.log(clone.a.b); // "alloy"
```
it works as (you maybe) expected

PR #477, opened 2025-03-17
PDCL-13324 Validation error for Variable data element when a working library is selected.

dompuiu:
## Description

There is an issue in Lens where it reloads the iframe when the `Save` button is pressed. While the iframe is reload, it calls also `getSettings` very early in the form loading process. That was crashing our view.

## Motivation and Context

The views are taking a few render cycles until the `values` property from Formik is no longer undefined. Notice the first 2 consoles have values key with undefined.

image

There will be also a fix in Lens to no longer reload the iframe when Save button is pressed. Not sure if it can be deployed this week due to Summit. Maybe it will be easier to release the extension this week. This is why I added this fix. 

Unfortunately it is very hard to test this fix using an automated test. I was not able to reproduce it in Sandbox.

## Types of changes

## Checklist:


carterworks (approved):
Regardless of the release process of the Lens fix, this is a reasonable change to make.

src/view/components/extensionView.jsx
```diff
@@ -135,7 +135,7 @@ const ExtensionView = ({
       try {
         return await viewRegistrationRef.current.getSettings({
```
jonsnyder, line 136: What if we did not call getSettings at all if the formik values are undefined?

dompuiu: @jonsnyder I changed the initial PR. Now getSettings will return `{}` when no Formik values are present.

carterworks (approved):

PR #490, opened 2025-07-10
Fix functional tests

carterworks:
## Description

There were two errors that were preventing the functional tests from running

### React serialization error

```
 ✖ renders with props: {}

   1) SyntaxError: unknown: Unexpected token, expected "," (1:39)
      
      > 1 | const func = (function run() { [native code] });
      |                                        ^

      Browser: Chrome 137.0.0.0 / Sequoia 15
```

`React.createElement()`, as used in `test/functional/specs/component/body.spec.mjs` and `test/functional/specs/component/heading.spec.mjs` was generating object that were not JSON serializable. When TestCafe tried to serialize them to pass them from the test to the on-page custom React renderer in `test/functional/helpers/components/fixture.jsx`, it threw an error because of the native code.

To fix this, we instead pass a pared-down version of the React element (just three keys: `type`, `props`, which has `props.children` and `key`) to the on-page custom React renderer in `test/functional/helpers/components/fixture.jsx`.

### CORS error

```
fixture.html:1 Access to script at
    'file://$REACTOR_EXTENSION_ALLOY_REPO_ROOT/componentFixtureDist/fixture.c38a08ef.js'
    from origin 'null' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes:
    chrome, chrome-extension, chrome-untrusted, data, http, https, isolated-app.
    fixture.html:112
    GET
    file://$REACTOR_EXTENSION_ALLOY_REPO_ROOT/componentFixtureDist/fixture.c38a08ef.js
    net::ERR_FAILED -->
```
TestCafe was loading the individual html pages as `file://` URLs, which are not allowed to load scripts due to CORS restrictions.

To fix this, we now create a dev server (using [`serve`](https://github.com/vercel/serve)) that serves the individual html pages as `http://` URLs and all the assets. This is done in `test/functional/helpers/components/createExtensionViewFixture.mjs`.
For component fixtures like `body.spec.mjs` and `heading.spec.mjs`, we now use tell parser to inline the script into the page. This is feasible since there is only one script and one page, instead of the many scripts and pages that the full extension uses in the extension view tests.

### Other

I also replaces [`npm-run-all`](https://github.com/mysticatea/npm-run-all) with [`concurrently`](https://github.com/open-cli-tools/concurrently) for running the build and test commands. `npm-run-all` doesn't support running commands that are not npm scripts and `concurrently` does.

## Types of changes

## Checklist:


dompuiu (approved):
So what changed? Did we update React? Why did it break at this moment?

carterworks:
> So what changed? Did we update React? Why did it break at this moment?

#487 is the first bad commit—functional tests work before that. 
I don't see anything [in the React changelog](https://github.com/facebook/react/blob/main/CHANGELOG.md) that indicates a change to `React.createElement`. Nor do I see anything in [the Babel 7.27.0 release notes](https://github.com/babel/babel/releases/tag/v7.27.0).
So I'm not sure what the issue is.

PR #498, opened 2025-07-22
Feature/use existing instance

ramboz:
## Description

This PR adds a flag to the configuration allowing to re-use a self-hosted `alloy.js` instance, so that the Launch container can extend it with data elements and rules. This is especially targeted towards the AEM EDS use cases, and to work towards better compatibility with [aem-martech](https://github.com/adobe-rnd/aem-martech/).

## Motivation and Context

In the context of EDS, where top CWV and LHS are of great importance, customers have started relying on direct `alloy.js` instrumentations. The [aem-martech](https://github.com/adobe-rnd/aem-martech/) was introduced to help with that, but we still have a gap in that we cannot easily integrate with a richer Launch container loaded async later in the page.

By allowing to re-use the existing `alloy.js` instance, we allow a seamless integration, and also align with the ACDL extension.

Screenshot 2025-07-22 at 15 32 24
Screenshot 2025-07-22 at 15 32 31

## Types of changes

## Checklist:

### DIFF
--- a/scripts/helpers/createExtensionManifest.mjs
+++ b/scripts/helpers/createExtensionManifest.mjs
@@ -262,7 +262,7 @@ const createExtensionManifest = ({ version }) => {
   const extensionManifest = {
     version,
     displayName: "Adobe Experience Platform Web SDK",
-    name: "adobe-alloy-eds",
+    name: "adobe-alloy",
     iconPath: "resources/images/icon.svg",
     exchangeUrl:
       "https://exchange.adobe.com/experiencecloud.details.106387.aep-web-sdk.html",
--- a/scripts/helpers/createExtensionManifest.mjs
+++ b/scripts/helpers/createExtensionManifest.mjs
@@ -262,7 +262,7 @@ const createExtensionManifest = ({ version }) => {
   const extensionManifest = {
     version,
     displayName: "Adobe Experience Platform Web SDK",
-    name: "adobe-alloy",
+    name: "adobe-alloy-eds",
     iconPath: "resources/images/icon.svg",
     exchangeUrl:
       "https://exchange.adobe.com/experiencecloud.details.106387.aep-web-sdk.html",
@@ -288,6 +288,9 @@ const createExtensionManifest = ({ version }) => {
                 name: {
                   type: "string",
                   pattern: "\\D+",
+                },
+                useExistingAlloy: {
+                  type: "boolean",
                 },
                 edgeConfigId: {
                   type: "string",
@@ -442,14 +445,40 @@ const createExtensionManifest = ({ version }) => {
                   additionalProperties: false,
                 },
               },
-              required: ["edgeConfigId", "name"],
               additionalProperties: false,
+              oneOf: [
+                {
+                  // If useExistingAlloy is true, only name is required.
+                  properties: {
+                    useExistingAlloy: {
+                      const: true,
+                    },
+                  },
+                  required: ["name"],
+                },
+                {
+                  // If useExistingAlloy is false or not defined,
+                  // then edgeConfigId is also required.
+                  properties: {
+                    useExistingAlloy: {
+                      const: false,
+                    },
+                  },
+                  required: ["name", "edgeConfigId"],
+                },
+              ],
             },
           },
           components: {
             type: "object",
             patternProperties: {
               ".*": { type: "boolean" },
+            },
+          },
+          hostedLibFiles: {
+            type: "array",
+            items: {
+              type: "string",
             },
           },
         },
--- a/src/lib/instanceManager/createInstanceManager.js
+++ b/src/lib/instanceManager/createInstanceManager.js
@@ -52,8 +52,26 @@ module.exports = ({
       stagingEdgeConfigId,
       developmentEdgeConfigId,
       onBeforeEventSend,
+      useExistingAlloy,
       ...options
     }) => {
+      if (useExistingAlloy) {
+        if (typeof window[name] === "function") {
+          instanceByName[name] = window[name];
+          if (!window.__alloyNS) {
+            window.__alloyNS = [];
+          }
+          if (window.__alloyNS.indexOf(name) === -1) {
+            window.__alloyNS.push(name);
+          }
+        } else {
+          turbine.logger.warn(
+            `Alloy instance "${name}" not found on window. Please ensure it is loaded before the Launch library.`,
+          );
+        }
+        return;
+      }
+
       const instance = createCustomInstance({ name, components });
       window[name] = instance;
       if (!window.__alloyNS) {
--- a/src/view/configuration/configuration.jsx
+++ b/src/view/configuration/configuration.jsx
@@ -15,15 +15,8 @@ import { object, array } from "yup";
 import { FieldArray, useField } from "formik";
 import {
   Button,
-  ButtonGroup,
-  Content,
-  DialogTrigger,
-  Dialog,
   Flex,
-  Heading as HeadingSlot,
   Item,
-  Divider,
-  Text,
   TabList,
   TabPanels,
   Tabs,
@@ -131,7 +124,7 @@ const getInitialValues = async ({ initInfo, context }) => {
 };
 
 const getSettings = async ({ values, initInfo }) => {
-  return {
+  const settings = {
     ...componentsBridge.getSettings({ values, initInfo }),
     instances: await Promise.all(
       values.instances.map((instanceValues) => {
@@ -143,6 +136,12 @@ const getSettings = async ({ values, initInfo }) => {
       }),
     ),
   };
+
+  if (values.instances.some((instance) => instance.useExistingAlloy)) {
+    settings.hostedLibFiles = [];
+  }
+
+  return settings;
 };
 
 const validationSchema = object().shape({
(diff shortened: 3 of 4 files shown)


src/view/configuration/configuration.jsx
```diff
@@ -143,6 +136,12 @@ const getSettings = async ({ values, initInfo }) => {
       }),
     ),
   };
+
+  if (values.instances.some((instance) => instance.useExistingAlloy)) {
+    settings.hostedLibFiles = [];
+  }
+
+  return settings;
```
jonsnyder, line 144: Near as I can tell, you are only ever setting this to []. If you want to use hostedLibFiles, you have to change the build and package process to generate another file. Seems we'll need to change this for every build though. We need to have a discussion if this will affect page performance for people not including their own Web SDK.

src/lib/instanceManager/createInstanceManager.js
```diff
@@ -52,8 +52,26 @@ module.exports = ({
...
+            window.__alloyNS.push(name);
+          }
+        } else {
+          turbine.logger.warn(
+            `Alloy instance "${name}" not found on window. Please ensure it is loaded before the Launch library.`,
+          );
+        }
+        return;
+      }
+
```
jonsnyder, line 74: If we don't find it initially, we should poll for it. Also ideally, we need to have some way to ensure that the instance has been configured already. (Perhaps use a monitoring hook?)

jonsnyder (changes requested):
Thanks for taking the time to submit a PR. Please schedule a time to meet with the team Serban Stancu, Carter McBride, Digant Tyagi, Joel Stachowitz, and myself and we can go over the requirements and design.

src/view/configuration/configuration.jsx
ramboz: My understanding, and I might be wrong here, is that setting it here basically overrides the backend logic that would otherwise embed `alloy.js` by default. So by forcibly emptying the array here when the toggle is set, I end up removing `alloy.js` from the build (since it's already self-hosted).

jonsnyder: hostedLibFiles is a property at the root level of the extension manifest, but you are changing the extension configuration settings here. Additionally, when you build the extension, the Web SDK files are packaged inside the library bundle, it is not a hosted lib file.

src/lib/instanceManager/createInstanceManager.js
ramboz: Sure, let me do something around that

dompuiu:
If we want to add this feature, we should probably add it in the same way we did for Adobe Analytics.

scripts/buildAlloy.cjs
```diff
@@ -142,46 +154,52 @@ alloyComponents.forEach((component) => {
...
-      return acc;
-    },
-    [],
-  );
+program.action(
+  async ({ inputFile, outputDir, shouldBuildAlloy, ...modules }) => {
+    if (!shouldBuildAlloy) {
+      console.log("shouldBuildAlloy is false, skipping custom alloy build.");
+      return;
+    }
```
ramboz, line 162: Same logic as before, just indented by 1 more "tab" and with a guard clause to skip the alloy build if we have any self-hosted alloy instance

scripts/helpers/createExtensionManifest.mjs
```diff
@@ -279,6 +279,9 @@ const createExtensionManifest = ({ version }) => {
         $schema: "http://json-schema.org/draft-04/schema#",
         type: "object",
         properties: {
+          shouldBuildAlloy: {
```
ramboz, line 282: Not surfaced in the UI and only used to set the build option

dompuiu:
> the analytics extension uses hostedLibFiles, but WebSDK is bundled, so I don't think we can necessarily align the build 
> processes easily here

Yes, this will be the most difficult part. We might be forced to no longer the bundle Web SDK and use the hostedLibFiles. It might affect other parts, I am not sure of the whole impact at this point.

dompuiu:
@ramboz I thought a little during the weekend about this feature. Here are a few thoughts, but you don't necessarily need to follow what I say.

- In the settings that will be saved to DB I would try to follow the approach we used in the Analytics extension. Even if we won't add all the options, Tags doesn't offer a settings migration option. In the future we might need to add more options and we should try to be flexible. Even if you add a checkbox in the UI, you could save it's value in a new property called `libraryCode`. This would look like:

```
{
  "libraryCode: {
      "type": "..." // Values of type could be managed and preinstalled.
  }
}
```
- I would not change the current build script for custom alloy. I would create a new script that would build an empty alloy and use the ENV variable inside the npm build script from [here](https://github.com/adobe/reactor-extension-alloy/blob/main/scripts/createExtensionPackage.mjs#L84).

You can have something similar with this pattern:

```
{
  "scripts": {
    "my-script": "[[ '$MY_VAR' == 'true' ]] && ./script1.sh || ./script2.sh"
  }
}
```

- The most difficult part will be the changes in this file: https://github.com/adobe/reactor-extension-alloy/blob/main/src/lib/instanceManager/createInstanceManager.js. It seems we are exposing createEventMergeId to other modules inside the extension and this will make the change tricky.

We can chat on a call for more details.

scripts/buildEmptyAlloy.cjs
```diff
@@ -0,0 +1,88 @@
...
+);
+
+program.action(({ outputDir }) => {
+  // Create stub content that uses the actual @adobe/alloy/utils
+  const proxyContent = `
+// Import the utility functions from @adobe/alloy/utils
+const { createEventMergeId, deepAssign } = require("@adobe/alloy/utils");
+
+// Create a proxy function for createCustomInstance that looks for window instances
+const createCustomInstance = (name) => {
```
ramboz, line 52: We mock the other 2 just for consistency, but they were only used to create managed instances, so should have 0 impact anyway

scripts/buildEmptyAlloy.cjs
```diff
@@ -0,0 +1,88 @@
...
+
+// Empty components object for preinstalled mode
+const components = {};
+
+// Export the implementations
+if (typeof module !== 'undefined' && module.exports) {
+  // CommonJS exports
+  module.exports = { createCustomInstance, components, createEventMergeId, deepAssign };
+} else {
+  // ES module exports
```
dompuiu, line 68: Tags doesn't support ES modules at this point. Only common js. I think you could remove the else branch.

src/lib/instanceManager/createInstanceManager.js
```diff
@@ -52,8 +54,68 @@ module.exports = ({
...
+            // eslint-disable-next-line no-unused-vars, unused-imports/no-unused-vars
+          } catch (err) {
+            turbine.logger.warn(
+              `Alloy instance "${name}" was not configured after 1000ms. Make sure the instance is configured before loading the Launch library.`,
+            );
+            return undefined;
+          }
+
+          // Latching: Overwrite the proxy in the map with the real instance.
+          instanceByName[name] = instance;
```
dompuiu, line 97: You set a proxy inside the instanceByName variable. But later in this file there are a few lines

```
      const instance = createCustomInstance({ name, components });
      .....
      instanceByName[name] = instance;
```

createCustomInstance returns something like this

```
     () => Promise.resolve({})
```

It seems that the return from createCustomInstance will overwrite the proxy. Is this what you want?

dompuiu:
When I run the sandbox I get this error inside the logs from the sandbox

```
[sandbox] An error was found in your extension.json: data/preprocessingVariables/8/default must be boolean
```

Do you see the same thing?

src/lib/instanceManager/createInstanceManager.js
```diff
@@ -54,30 +57,41 @@ module.exports = ({
...
       }
-      const environment = turbine.environment && turbine.environment.stage;
 
-      const computedEdgeConfigId =
-        (environment === "development" && developmentEdgeConfigId) ||
-        (environment === "staging" && stagingEdgeConfigId) ||
-        edgeConfigId;
+      // Only configure if NOT using preinstalled mode
+      // In preinstalled mode, the external instance is already configured by user code
+      if (!isPreinstalled) {
```
ramboz: Just wraps all the existing logic in a new conditional

src/lib/utils/createPreinstalledProxy.js
```diff
@@ -0,0 +1,118 @@
...
+ * @param {String} name - The instance name to look for on window
+ * @param {Object} [options={}] - Configuration options
+ * @param {Number} [options.timeout=1000] - Max time to wait for instance (ms)
+ * @param {Number} [options.interval=100] - Polling interval (ms)
+ * @param {Number} [options.configTimeout=5000] - Max time to wait for getLibraryInfo (ms)
+ * @param {Function} [options.onWarn] - Warning callback (msg) => void
+ * @param {Function} [options.onError] - Error callback (msg) => void
+ * @return {Function} - Proxy function that forwards to real instance
+ */
+export const createPreinstalledProxy = (name, options = {}) => {
```
ramboz: This creates a "proxy" instance that will poll for the real preinstalled instance on the `window` object.
The goal is to address possible race conditions between the loading of the pre installed instance against the launch container.

src/view/configuration/configuration.jsx
```diff
@@ -110,6 +112,46 @@ const getInitialInstanceValues = getMergedBridgeMethod(
 );
 const getInstanceSettings = getMergedBridgeMethod("getInstanceSettings");
 
+/**
+ * Creates the validation schema based on library type
+ */
+const createValidationSchema = () => {
```
ramboz: Dynamically evaluates the validation schema based on the `libraryCode.type` state

src/view/configuration/configurationView.jsx
```diff
@@ -348,25 +413,49 @@ const Configuration = ({ initInfo, context }) => {
...
+          label="Alloy library configuration"
+          description="Choose how the Alloy library should be loaded"
+          orientation="horizontal"
+        >
+          <Radio value="managed">Managed by Launch (default)</Radio>
+          <Radio value="preinstalled">
+            Use existing alloy.js instance (self-hosted)
+          </Radio>
+        </FormikRadioGroup>
+      </View>
```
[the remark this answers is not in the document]
jonsnyder: Maybe "Web SDK build options"

src/lib/utils/deepAssign.js
ramboz: A deep assign implementation that is bundled int he empty alloy instance

src/view/dataElements/eventMergeId.jsx
ramboz: Warns that event merge is not available with preinstalled mode

test/integration/helpers/ui.js
```diff
@@ -33,7 +33,7 @@ export const expandAccordion = async (name) => {
 };
 
 export const toggleComponent = async (component) => {
-  await expandAccordion("Custom build components");
+  await expandAccordion("Web SDK build options");
```
ramboz: Renamed the custom build section as per comment

src/view/components/requiredComponent.jsx
```diff
@@ -33,11 +34,57 @@ const RequiredComponent = ({
...
+      // This is returned when in preinstalled mode and the item is new.
+      return (
+        <InlineAlert
+          variant="negative"
+          width="size-5000"
+          data-test-id="preinstalledModeError"
+        >
+          <Heading>Not available in preinstalled mode</Heading>
+          <Content>
+            You cannot create {title} when using the preinstalled library type.
```
jonsnyder: change the wording here too.

PR #520, opened 2025-08-28
Exclude Advertising and Push notification components by default when extension is upgraded

jonsnyder:
## Description

* For customers that are upgrading their Web SDK extension in Tags, we only want their original set of components to be used.
* Before this change we were storing just the components to exclude.
* This change will make any new components excluded by default when building and when showing the UI.
* I created an allowlist of "original" components that default to true, and then any new components default to false.
* 

## Types of changes

## Checklist:

### DIFF
--- a/scripts/helpers/createExtensionManifest.mjs
+++ b/scripts/helpers/createExtensionManifest.mjs
@@ -12,11 +12,22 @@ governing permissions and limitations under the License.
 
 import getAlloyComponents from "./getAlloyComponents.mjs";
 
+const ORIGINAL_DEFAULT_COMPONENTS = new Set([
+  "activityCollector",
+  "audiences",
+  "consent",
+  "eventMerge",
+  "mediaAnalyticsBridge",
+  "personalization",
+  "rulesEngine",
+  "streamingMedia",
+]);
+
 const createPreprocessingVariables = () =>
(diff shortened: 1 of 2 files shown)


scripts/helpers/createExtensionManifest.mjs
```diff
@@ -12,11 +12,22 @@ governing permissions and limitations under the License.
 
 import getAlloyComponents from "./getAlloyComponents.mjs";
 
+const ORIGINAL_DEFAULT_COMPONENTS = new Set([
```
dompuiu, line 15: Maybe we should name this just as DEFAULT_COMPONENTS. If we add a new component later and we want to enable it by default I think we just need to add it to this Set.

src/view/configuration/componentsSection.jsx
```diff
@@ -89,27 +89,26 @@ export const bridge = {
...
-        return acc;
-      }, {});
-
-    if (Object.keys(excludedComponents).length > 0) {
-      return { components: excludedComponents };
-    }
-
-    return {};
+    // return all values regardless of whether they are set or not
+    return { components };
```
dompuiu, line 111: As an optimization maybe we should not save the component to the settings if the value matches the default value.

dompuiu:
This PR is replaced by #521 .

PR #538, opened 2025-09-29
Fixed autoPopulationAlert to evaluate node's absolute path

Spencer-Smith:
## Description

The variable editor on the UpdateVariable action page currently has a small visual bug: after adding an item to an array, the nested object is evaluated without its absolute path when determining the autoPopulationSource. So if a field in the new nested object shares a name with a top-level auto-populating field (eg "_id"), that field will incorrectly be marked as auto-populating.

This PR addresses that issue by passing the node path to the function responsible for determining the initial state of the object when it is created in that view, so that it correctly determines the autoPopulationSource for each field.

## Related Issue

[Jira issue](https://jira.corp.adobe.com/browse/PDCL-6667)

## Motivation and Context

This change fixes a bug that would misinform users about how data is populated in certain fields.

(See jira issue for example of incorrect auto-population notice)

## Types of changes

## Checklist:

- [ ] ~My change requires a change to the documentation.~ *No*

### DIFF
--- a/src/view/components/objectEditor/arrayEdit.jsx
+++ b/src/view/components/objectEditor/arrayEdit.jsx
@@ -107,10 +107,15 @@ const PartsPopulationStrategyForm = ({
                   defaultValue = "";
                 }
 
+                const itemNodePath = fieldName
+                  ? `${fieldName}.items.${items.length}`
+                  : `items.${items.length}`;
+
                 const itemFormStateNode = getInitialFormState({
                   schema: itemSchema,
                   value: defaultValue,
                   updateMode,
+                  nodePath: itemNodePath,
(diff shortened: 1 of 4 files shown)


jonsnyder (approved):
Great work!

Spencer-Smith:
> What's the path for testing this?

I was unsure on this. I wrote some unit tests but they didn't seem to match patterns in the repo for unit tests, especially because there's not much "logic" to unit test. I looked at some of the functional tests in the repo, there are some for the view (Update Action), but not for a slightly lower level (ie at the "arrayEdit" or "getInitialState" level), so I wasn't quite sure what the recommendation is?

PR #539, opened 2025-09-29
Fix the get settings method from the `Send Media` action.

dompuiu:
## Description

From a report on Slack. We changed half a year ago the `Send Media Event` action to show a warning when the `Streaming Media` module is disabled. When this changed was made, we broke the getSetting.

I would have added a test, but at this moment the testcafe tests are not running anymore on my machine.

## Types of changes

## Checklist:

### DIFF
--- a/src/view/actions/sendStreamingMediaEvent.jsx
+++ b/src/view/actions/sendStreamingMediaEvent.jsx
@@ -237,6 +237,7 @@ const customMetadataSection = section(
     objectArray(
       {
         name: "customMetadata",
+        ariaLabel: "Custom metadata",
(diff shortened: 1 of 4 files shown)


src/view/actions/sendStreamingMediaEvent.jsx
```diff
@@ -237,6 +237,7 @@ const customMetadataSection = section(
     objectArray(
       {
         name: "customMetadata",
+        ariaLabel: "Custom metadata",
```
dompuiu, line 240: Without this, the log was flooded with warnings for every key stroke.

